PR #17, opened 2018-10-22
Chore: clean up Dockerfile

gruebel:
Changed `cp` to `mv`, because it is faster.

Removed `chmod`, because is already executable after the build.

Removed `mkdir`, because the directory will be created automatically when using `COPY`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,23 @@
 FROM golang:latest as builder
+
 RUN wget http://geolite.maxmind.com/download/geoip/database/GeoLite2-Country.tar.gz -O /tmp/GeoLite2-Country.tar.gz && \
     tar zxvf /tmp/GeoLite2-Country.tar.gz -C /tmp && \
-    cp /tmp/GeoLite2-Country_*/GeoLite2-Country.mmdb /Country.mmdb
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,17 +1,23 @@
 FROM golang:latest as builder
+
 RUN wget http://geolite.maxmind.com/download/geoip/database/GeoLite2-Country.tar.gz -O /tmp/GeoLite2-Country.tar.gz && \
     tar zxvf /tmp/GeoLite2-Country.tar.gz -C /tmp && \
-    cp /tmp/GeoLite2-Country_*/GeoLite2-Country.mmdb /Country.mmdb
+    mv /tmp/GeoLite2-Country_*/GeoLite2-Country.mmdb /Country.mmdb
```
Dreamacro: Golang is only a builder here, right? There seems to be no difference between `mv` and `cp`. So what is the meaning of the changes here?

gruebel: In general moving a file on the same partition is faster than copying it, because `mv` just remaps the inode.

gruebel: Also true :D was not sure, if I should change it or not. But happy to be able to contribute a bit :)

PR #42, opened 2018-11-21
support socks5 username/password authentication

felix021:
add username/password support for socks5, tested with gost type "socks5+tls://username:password@ip:port"


felix021:
found this feature already merged in https://github.com/Dreamacro/clash/pull/34

PR #44, opened 2018-11-22
fix: path.IsAbs is not implemented for windows, use filepath.IsAbs instead

felix021:
On windows: `path.IsAbs("D:/clash")` returns false.

Refer to: https://github.com/golang/go/issues/1483

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,6 +6,7 @@ import (
 	"syscall"
 	"flag"
 	"path"
+	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


Dreamacro:
already fixed in 91e35f2f6a4a006e87682f91cf5e151b86c57fc0

felix021:
great! with those fixes, it satisfied all my needs now, thanks for the wonderful project.

PR #63, opened 2018-12-08
FreeBSD compatibility patch

yurivict:


Dreamacro:
Have you tested it? I don't have FreeBSD 😅

Dreamacro:
BTW, Is your patch different from https://github.com/Dreamacro/clash/blob/master/proxy/redir/tcp_linux_other.go#L10 ?

yurivict:
It builds fine with the patch, but I didn't run it.

> BTW, Is your patch different from

About the same.

PR #75, opened 2018-12-21
Fix: allow access to external-ui without authentication

songchenwen:


Dreamacro:
please use `gofmt` to format the code and `r.Use(cors.Handler)` is useless in fileserver.

songchenwen:
I have updated this PR.

PR #98, opened 2019-01-19
Update README.md

kongminhao:
fix cipher name spell case error.


Dreamacro:
The cipher name is case insensitive

kongminhao:
ok, thanks.

PR #127, opened 2019-02-28
support TCP DNS & log the DNS parsing error

### LINKED ISSUE #126: Improvement about DNS config(TLS and TCP dns)
DNS part of clash config is confusing, I'm planning to submit a PR to fix  two problems.

### When using TLS DNS server, the `port` field cannot be omitted

For example, If I'm using following configure
```
dns:
  enable: true # set true to enable dns (default is false)
  ipv6: false # default is false
  listen: 0.0.0.0:53
  # enhanced-mode: redir-host
  nameserver:
    - tls://1.1.1.1
```

Read the code
https://github.com/Dreamacro/clash/blob/e7997a035bfcbc943c0b71de2b0367a8183e8676/config/config.go#L373-L382
When using `tls://1.1.1.1:853`, SplitHostPortwill generate an error  `address tls://1.1.1.1:853: too many colons in address`

But When using `tls://1.1.1.1`, 
The `hostname` equals `tls` and the port equals `//1.1.1.1` without error generated
Red the code
https://github.com/Dreamacro/clash/blob/e7997a035bfcbc943c0b71de2b0367a8183e8676/config/config.go#L390-L396
Then, **a nameserver config with `Host` equal "tls" is created without any error log**

So, the code to set a default `853` port  is useless
https://github.com/Dreamacro/clash/blob/e7997a035bfcbc943c0b71de2b0367a8183e8676/config/config.go#L407

### Clash doesn't support TCP DNS server

For now,  Clash doesn't support UDP. So if I want to use DNS servers via proxy, I must choose TCP (or tcp-tls) server. 

I tried config like `tcp://1.1.1.1`, It didn't work and **didn't genrate any error log too**. The same reason as above

comzyh:
- add TCP DNS support in DNS config
- log the DNS config parsing error 

fix #126

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -386,28 +386,26 @@ func parseNameServer(servers []string) ([]dns.NameServer, error) {
 	nameservers := []dns.NameServer{}
 
 	for idx, server := range servers {
+		// parse without scheme .e.g 8.8.8.8:53
+		if !strings.Contains(server, "://") {
+			server = "udp://" + server
+		}
 		u, err := url.Parse(server)
 		if err != nil {
 			return nil, fmt.Errorf("DNS NameServer[%d] format error: %s", idx, err.Error())
 		}
 
-		// parse without scheme .e.g 8.8.8.8:53, u will have empty `Host`
-		if u.Host == "" {
-			u.Host = u.Path
-			u.Path = ""
-		}
-
-		var host, dnsNet string
+		var host, dnsNetType string
 		switch u.Scheme {
-		case "":
+		case "udp":
 			host, err = hostWithDefaultPort(u.Host, "53")
-			dnsNet = "" // UDP
+			dnsNetType = "" // UDP
 		case "tcp":
 			host, err = hostWithDefaultPort(u.Host, "53")
-			dnsNet = "tcp" // TCP
+			dnsNetType = "tcp" // TCP
 		case "tls":
 			host, err = hostWithDefaultPort(u.Host, "853")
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -386,34 +386,43 @@ func parseNameServer(servers []string) ([]dns.NameServer, error) {
...
-		// parse without scheme .e.g 8.8.8.8:53
-		if host, err := hostWithDefaultPort(server, "53"); err == nil {
-			nameservers = append(
-				nameservers,
-				dns.NameServer{Addr: host},
-			)
-			continue
-		}
-
 		u, err := url.Parse(server)
```
Dreamacro: The `url.Parse` can't parse `8.8.8.8:53`, it would return an error 🤓

config/config.go
```diff
@@ -430,10 +439,14 @@ func parseDNS(cfg rawDNS) (*DNS, error) {
 
 	if nameserver, err := parseNameServer(cfg.NameServer); err == nil {
 		dnsCfg.NameServer = nameserver
+	} else {
+		return nil, err
```
beyondkmp:
```go
if dnsCfg.NameServer, err := parseNameServer(cfg.NameServer); err != nil{
    return nil , err
}
```
这样编码是不是更好点？

comzyh: 好

config/config.go
comzyh: OK, fixed

Dreamacro:
close #126

PR #105, opened 2019-01-22
 Feature: DNS fallback-ip configuration

songchenwen:
This allows us to tell clash to give up poisoned DNS results like 243.185.187.39

Issue #95, opened 2019-01-15
Some poisoning IP can not use fallback DNS 


comzyh:
For example, the following IP is poisoning IP, but their country code in GEOIP is empty, so clash will not use fallback DNS for them.

> 243.185.187.39
> 249.129.46.48
> 253.157.14.165

https://github.com/Dreamacro/clash/blob/49635eab6cbc5c667ea1d015d4f3c06851f53a01/dns/client.go#L134-L141

Why does clash use fallback when `record.Country.IsoCode == ""` ?

Dreamacro:
> 1. when clash is restarted, internal DNS cache of clash will go disappear

Is that a problem?

> 2. clash DNS server doesn‘t apply DNS TTL countdown, so there is a chance that a downstream DNS 

It will be fixed later.

> 3. multiple different domains may have same “poisoned” IP, so only one of the domains can establish a correct connection.

I don't understand what it means.

comzyh:
github.com and google.com my have identical ip from poisoned dns cache.

Dreamacro <notifications@github.com> 于 2019年1月19日周六 19:53写道：

BirkhoffLee:
I am having exact same issue that Clash
is getting poisoned IP addresses. This makes redir and dns feature literally unusable. It happens on many domains including google.com,duckduckgo.com, etc

comzyh:
> > 1. when clash is restarted, internal DNS cache of clash will go disappear
> 
> Is that a problem?

I think No.1 is a problem, for Android devices,  shutting down and restarting is not a rare case.

comzyh:
Update， I found that some DNS poisoning is using `0.0.0.0` now, and multiple IP direct to the same `0.0.0.0`,  and #105  works well for me now.

```
dig www.google.com @10.10.0.21
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 60429
;; Flags: qr rd ra; QUERY: 1; ANSWER: 1; AUTHORITY: 4; ADDITIONAL: 4

;; QUESTION SECTION:
;; www.google.com. IN A

;; ANSWER SECTION:
www.google.com. 13331 IN A 0.0.0.0

;; AUTHORITY SECTION:
google.com. 1548 IN NS ns3.google.com.
google.com. 1548 IN NS ns4.google.com.
google.com. 1548 IN NS ns1.google.com.
google.com. 1548 IN NS ns2.google.com.

;; ADDITIONAL SECTION:
[...]

PR #142, opened 2019-03-18
using fallback dns for reserved address

Fndroid:
```
The IP address '243.185.187.39' is a reserved IP address (private, multicast, etc.).
```
This IP will make Chrome shows Error like https://github.com/Dreamacro/clash/issues/103

So when got a reversed IP that isoCode is empty, use fallback dns instead.


comzyh:
Same problem #95
And there is already a fix #105

10.0.0.0/8 or 192.168.0.1 are also reserved IPs. 

If upstream reply such IPs, clash should not use fallback DNS.

PR #144, opened 2019-03-18
Fix alive checking

comzyh:
The `fallback` group is not working now.
No error in `Dial` doesn't mean proxy is alive. Actually, if ss server is reachable, you can always dial it without any error. 

Alive status should depend on the result of URLTest.

### DIFF
--- a/adapters/outbound/base.go
+++ b/adapters/outbound/base.go
@@ -43,7 +43,7 @@ func (p *Proxy) Alive() bool {
 
 func (p *Proxy) Dial(metadata *C.Metadata) (net.Conn, error) {
 	conn, err := p.ProxyAdapter.Dial(metadata)
-	p.alive = err == nil
+	p.alive = p.alive && (err == nil)
 	return conn, err
 }
 
@@ -89,6 +89,7 @@ func (p *Proxy) URLTest(url string) (t uint16, err error) {
 		if err == nil {
 			record.Delay = t
 		}
+		p.alive = err == nil
 		p.history.Put(record)
 		if p.history.Len() > 10 {
 			p.history.Pop()


Dreamacro:
The `Dial` is Similar to `net.Dial` but a little difference. It contains handshake with proxy node.

comzyh:
Sure, but what does 'alive' mean?

If clash can establish TCP connection with proxy server, but can't connect any website via that proxy(incorrect password, etc).

Is such proxy consider as 'alive' ?

Dreamacro:
`alive` means that the proxy can throughput data. if `ProxyAdapter.Dial` return an error, the proxy node "fall asleep". When you call "delay test" or autotest in the next cycle, the proxy node may "wake up" when `Dial` success, then throughput data as usual.

comzyh:
OK, here is the problem.
If you set an incorrect password for a shadowsocks node, that node can not throughput data. 
But `Dial` will not return any error.

Dreamacro:
The correctness of shadowsocks password is not the goal of `alive`, you can checkout correctness 
 with "delay test". The more common situation is temporary disconnection due to network fluctuation. So I think `Dial` is enough. Do you have any different opinions on this?

comzyh:
The incorrect password is just one of the examples.

1. Many users are using so-called 'Relay' server. In other words, a reverse proxy of a remote proxy server. Connectivity to relay server doesn't mean connectivity to the relay proxy server.
2. Server overload. Due to the complex structure of the proxy server, an overloaded server may accept TCP connection, but timeout when real traffic is transmitted.

 

> you can checkout correctness with "delay test".

The problem is that the `fallback` group is choosing the backend according to `Alive` status.
https://github.com/Dreamacro/clash/blob/7a9d986ff38c356ef78c720b3cb5a597bd5ca3c3/adapters/outbound/fallback.go#L68-L75

If `alive` just check TCP connectivity to the remote proxy server, ignore the connectivity to the real server.  **Why** does clash need a `url` field in `fallback` outbound?

https://github.com/Dreamacro/clash/blob/7a9d986ff38c356ef78c720b3cb5a597bd5ca3c3/adapters/outbound/fallback.go#L21-L24

Dreamacro:
Take a look f99da37 and have a try

comzyh:
I'm sorry I can't tell any difference in f99da37. 
The only meaningful change is that if `proxy.alive` equal `false` then `LastDelay()`  will return `0xffff`. 
Other changes are replacing `0 - 1` with `0xffff`and replacing `p.ProxyAdapter.Dial` with `p.Dial`. Can you explain these? I think each of these two is only another expression of original logic.

And I have tested f99da37, nothing changes. `Fallback` group still use the dead proxy (dead proxy is proxy can establish a TCP connection with the proxy server but show `Unavailable` or `Timeout` in dashboard)

Dreamacro:
目前所有的 "proxy" 或者 "proxy group" 在之前的实现中，是作为一个 `ProxyAdapter`，外部再包上一个 `Proxy` (我也不想这么做，golang 只能这样了)。之前的 `URLTest` 里使用的是 `ProxyAdapter.Dial`，它并不会触发 alive 的改变。修改 `LastDelay` 是因为这导致了 `url-test` 失效的原因。

`Unavailable` 意味着代理信息错误，导致远端请求失败。`Timeout` 意味着在指定时间 (Dashboard 是 5s) 内没有返回信息，但不代表这个节点不可用。目前也只能在 URLTest 再 set alive 来保证了。

adapters/outbound/base.go
```diff
@@ -43,7 +43,7 @@ func (p *Proxy) Alive() bool {
 
 func (p *Proxy) Dial(metadata *C.Metadata) (net.Conn, error) {
 	conn, err := p.ProxyAdapter.Dial(metadata)
-	p.alive = err == nil
+	p.alive = p.alive && (err == nil)
```
Dreamacro: 这句修改没有意义

comzyh:
有的，我们把联通情况分为两种:
1. 服务器可以TCP握手
2. URLTest可以通过

1是2的前提，如果服务器不能TCP握手则URLTest一定不能通过。

因为Dial是一个经常操作，能够在一次TCP握手失败后立即将Alive设置成False，有响应速度快的优势。

但是TCP握手成功并不能表示URLTest可以通过，所以用一个与逻辑，如果本身Alive是flase，不能让Dial将alive设置成True.


Dreamacro <notifications@github.com> 于 2019年3月23日周六 19:16写道：

comzyh:
resolved in 18f885a92ab62524d5ff96852b7897fc7cbdbe3a

PR #151, opened 2019-03-24
Fix: use cas atomic

beyondkmp:


comzyh:
The original logic is:

If `u.once` + 1 != 1, then return

The newer logic is:

if `u.once` equals `0`, then swap it with `1`, and **return**


Correct me if i am wrong.

beyondkmp:
@comzyh you are right. The correct way should be:
```go
if !atomic.CompareAndSwapInt32(&u.once, 0, 1) {
    return
}
```

PR #155, opened 2019-04-01
Add GitBook link in README.md

BirkhoffLee:
I had to ask people in the Telegram group to check the API docs, so I think adding this to the README will be very useful.


ojbkfoo:
我也是找了好久。。。真的应该在readme里加个链接

PR #162, opened 2019-04-11
 New feature: add support for SO_MARK for Linux socket options

gaorocks:
Thanks to Netfilter TCP redirect, Clash is able to be setup as a transparent proxy.

Let's say we setup Clash on an OpenWrt gateway, redirecting TCP traffic to Clash by Netfilter/iptables, all devices connected to the gateway are proxied automatically and transparently.

Regarding the last mile, the gateway itself, we have to deal with chain OUTPUT just like chain PREROUTING (Inspired by this [article](https://toutyrater.github.io/app/transparent_proxy.html)). At the same time, outbound traffic processed by Clash needs to be bypassed. To bypass specific traffic, we could mark each packet by setting specified SO_MARK socket options within Clash and filter by SO_MARK at chain OUTPUT. Then problem solved.

Note: SO_MARK only works on Linux, but it is enough since most router/gateway operating systems are based on Linux.

### clash config.yml

```
socket-mark: 0xffff
```

### iptables NAT rules

```
iptables -t nat -N CLASH
iptables -t nat -A CLASH -d 127.0.0.1,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16 -j RETURN
iptables -t nat -A CLASH -p tcp -m mark --mark 0xffff -j RETURN
iptables -t nat -A CLASH -p tcp -j REDIRECT --to-ports 7892
iptables -t nat -A PREROUTING -p tcp -j CLASH
iptables -t nat -A OUTPUT -p tcp -j CLASH
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,10 @@ external-controller: 127.0.0.1:9090
 # Secret for RESTful API (Optional)
 # secret: ""
 
+# SO_MARK for Linux socket options which may help filter and bypass outbound packet processed by clash
+# in order to build a transparent proxy together with Netfilter redirect
+# socket-mark: 0xffff
+
 # dns:
   # enable: true # set true to enable dns (default is false)
   # ipv6: false # default is false
--- a/adapters/outbound/direct.go
+++ b/adapters/outbound/direct.go
@@ -16,7 +16,7 @@ func (d *Direct) Dial(metadata *C.Metadata) (net.Conn, error) {
 		address = net.JoinHostPort(metadata.IP.String(), metadata.Port)
 	}
 
-	c, err := net.DialTimeout("tcp", address, tcpTimeout)
+	c, err := dialTimeout("tcp", address, tcpTimeout)
 	if err != nil {
 		return nil, err
 	}
--- a/adapters/outbound/http.go
+++ b/adapters/outbound/http.go
@@ -36,7 +36,7 @@ type HttpOption struct {
 }
 
 func (h *Http) Dial(metadata *C.Metadata) (net.Conn, error) {
-	c, err := net.DialTimeout("tcp", h.addr, tcpTimeout)
+	c, err := dialTimeout("tcp", h.addr, tcpTimeout)
 	if err == nil && h.tls {
 		cc := tls.Client(c, h.tlsConfig)
 		err = cc.Handshake()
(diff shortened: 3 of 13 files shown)


Kr328:
为什么不试试 使用 uid 或者 cgroup 标记 clash 的流量 以达到直连的效果

gaorocks:
> 为什么不试试 使用 uid 或者 cgroup 标记 clash 的流量 以达到直连的效果

好主意 已用 uid 实现 这样更优雅一点 感谢大佬提醒

```
LAN_IP_CIDR=127.0.0.1,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16
CLASH_REDIR_PORT=7892
CLASH_USER=clash

iptables -t nat -N CLASH
iptables -t nat -A CLASH -d $LAN_IP_CIDR -j RETURN -m comment --comment 'bypass traffic to LAN'
iptables -t nat -A CLASH -p tcp -j REDIRECT --to-ports $CLASH_REDIR_PORT -m comment --comment 'redirec rest traffic to clash'

iptables -t nat -A PREROUTING -p tcp -j CLASH -m comment --comment 'PREROUTING traffic to be handled by clash'

iptables -t nat -A OUTPUT -m owner --uid-owner $CLASH_USER -j RETURN -m comment --comment 'bypass traffic processed by clash'
iptables -t nat -A OUTPUT -p tcp -j CLASH -m comment --comment 'OUTPUT traffic to be handled by clash'
```

@Dreamacro You may proceed to close this PR now since there's a better approach mentioned above. Thanks all the same.

myfingerhurt:
@gaorocks, thank you for your PR. I need this so_mark so much, this topic helps me a lot.
Here is some improvements I have found. And I still believe it's a good idea to carry on this PR. some routers may need this PR, add another user for those routers is not a feasible option sometimes. 
 
Without root privilege you would encounter those errors.
binding low port gets denied, and opening UDP port fails.
```
ERRO[0000] Start DNS server error: listen udp 127.0.0.1:53: bind: permission denied
WARN[0000] Failed to start Redir UDP Listener: operation not permitted
```

Use this to start a service as non-root but bind low ports. 
`setcap 'cap_net_admin,cap_net_bind_service=+ep' /usr/bin/clash`

Run clash as clash user that has no password.  Everything should be OK now.
adduser clash --shell=/bin/false --system --no-create-home
`sudo -H -u clash clash -d /path2config/.clash`

PR #176, opened 2019-05-06
Add fake-ip-range in example configuration

BirkhoffLee:


Dreamacro:
`fake-ip-range` default value is `198.18.0.1/16`, `198.18.0.1/31` seems no working due to few valid IP

Dreamacro:
close because of README.md is already update.

PR #186, opened 2019-05-18
fix: DNS server not recreate correctly

Fndroid:


Dreamacro:
It's been tested, right?

PR #187, opened 2019-05-18
fix fake ip dns ttl

Kr328:
https://github.com/Dreamacro/clash/commit/0d4a999707da2ec0c989aa3fe2b8faff5950340c

dns return origin ttl when hit cache


Dreamacro:
Please remove https://github.com/Dreamacro/clash/blob/03626452099aad81fe7ece481c4ea9f8a8197f8a/dns/server.go#L75-L76

Kr328:
updated

PR #188, opened 2019-05-18
 Feature: save fakeip dns records in file

beyondkmp:


Dreamacro:
It seems to implement cache after `events` may easier than now.

beyondkmp:
Yes, I have some tricks to implement saving fakeips to file.  It's just used  in Raspberry pie for me. I pull request for marking this demand.

common/cache/cache.go
```diff
@@ -59,6 +62,57 @@ func (c *cache) GetWithExpire(key interface{}) (payload interface{}, expired tim
 	return elm.Payload, elm.Expired
 }
 
+func (c *cache) Save(filepath string) error {
+	f, err := os.Create(filepath)
+	defer f.Close()
+
```
recall704: you need handle err before `defer f.Close()`

PR #190, opened 2019-05-21
feature: add arguments to specify config file name in config folder

SpongeNobody:
Add new argument (-c) to specify config file name in config folder.

Could be handy when subscribing from multiple service.

### DIFF
--- a/constant/path.go
+++ b/constant/path.go
@@ -12,7 +12,8 @@ const Name = "clash"
 var Path *path
 
 type path struct {
-	homedir string
+	homedir  string
+	confname string
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


Dreamacro:
Only changing the filename is not flexible enough and I prefer the way like `-d`.

```
> echo $PWD
/tmp
> clash -d . -c another.yml
# Directory is /tmp and config file is /tmp/another.yml
```

```
> clash -d /tmp -c /opt/another.yml
# Directory is /tmp and config file is /opt/another.yml
```

BTW, clash RESTful API provides an API for changing the config file without restart clash (https://clash.gitbook.io/doc/restful-api/config)

SpongeNobody:
Hmmm I didn't notice clash already support that via API. Closing this PR. Sorry.

PR #203, opened 2019-06-14
Makefile: add android build target

Kr328:
Add android build target to Makefile

on linux
```bash
ANDROID_NDK=/path/to/android-ndk make android-arm64-v8a
```

on Windows 64-bit
```bash
ANDROID_NDK=/path/to/android-ndk ANDROID_NDK_HOST=windows-x86_64 make android-arm64-v8a
```


Dreamacro:
I wouldn't like to depend on android sdk....

Kr328:
@Dreamacro 
Only require Android NDK when building android version (

Dreamacro:
So the best way is don't include android release to avoid clash become bloated. 🤔

Dreamacro:
Unfortunately, I have to close this PR 🌚

PR #210, opened 2019-06-20
Fix: checking target is valid in rules

beyondkmp:
enhancement : #207

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -377,7 +377,7 @@ func parseRules(cfg *rawConfig, proxies map[string]C.Proxy) ([]C.Rule, error) {
 		}
 
 		if _, ok := proxies[target]; !ok {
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -376,6 +376,10 @@ func parseRules(cfg *rawConfig) ([]C.Rule, error) {
 			return nil, fmt.Errorf("Rules[%d] [%s] error: format invalid", idx, line)
 		}
 
+		if _, ok := proxies[target]; !ok {
+			return nil, fmt.Errorf("Rules[%d] [%s] error: proxy target invalid", idx, line)
```
Dreamacro: "invalid" --> "not found" may better?

beyondkmp: ok

PR #216, opened 2019-06-25
Feature: local socks5/http(s) server auth

popu125:
为本地 socks5 服务器添加认证功能，便于暴露到 lan 时的访问控制。


recall704:
good job！

can you add a auth.txt example  file ?

popu125:
> good job！
> 
> can you add a auth.txt example file ?

The format is `user:pass`, one record per line.

Dreamacro:
I have a quick look and find some issue and doubt

* About `import`

clash follows a simple rule in `import`

```go
import (
    // go pkg

    // self pkg

    // third party pkg
)
```

* About `auth.txt`

why not use `yaml` like:

```yaml
auth-source:
  - "user:pass"
  - "user2:pass"
```

* About inject

I prefer not to inject Authenticator in `NewSocksProxy` due to maybe we need hot replace another Authenticator.

* About HTTP

It seems like `auth-source` share Authenticator with other inbound proxies like HTTP, but it's not implemented yet.

recall704:
> > good job！
> > can you add a auth.txt example file ?
> 
> The format is `user:pass`, one record per line.

I mean you need update doc to explain how to use it for the normal user.

popu125:
> I have a quick look and find some issue or doubt
> 
> * About `import`
> 
> clash follows a simple rule in `import`
> 
> ```go
> import (
>     // go pkg
> 
>     // self pkg
> 
>     // third party pkg
> )
> ```
> 
> refer to https://github.com/Dreamacro/clash/blob/dev/config/config.go
> 
> * About `auth.txt`
> 
> why not use `yaml` like:
> 
> ```yaml
> auth-source:
>   - "user:pass"
>   - "user2:pass"
> ```
> 
> * About inject
> 
> I prefer not to inject Authenticator in `NewSocksProxy` due to maybe we need hot replace another Authenticator.
> 
> * About HTTP
> 
> It seems like `auth-source` share Authenticator with other inbound proxies like HTTP, but it's not implemented yet.

 - About `auth.txt`

When I started this work, I was just get away from the snail007/goproxy so I thought save login informations in a file outside of all-in-one config file is easier to modify by automatic scripts. 

The code was wanted to be multi-source hot reload, but the DB connection looks useless. So i kept the design and seek for your opinion here.

Or I'll put them in config.

 - About inject 

It looks that you use `ReCreateSocks` to reload the config so I think call the `proxy.SetAuthenticator` may be a acceptable solution, looks not?

 - About HTTP

I was wanted to finish the HTTP auth at the same time. But the headache makes me forgot it. I'll finish it this evening.

popu125:
> > > good job！
> > > can you add a auth.txt example file ?
> > 
> > 
> > The format is `user:pass`, one record per line.
> 
> I mean you need update doc to explain how to use it for the normal user.

I'll do it when the final way to save login info was completely implemented.

Dreamacro:
> so I thought save login informations in a file outside of all-in-one config file is easier to modify by automatic scripts.

Unless you are using the shell script, I don't think modifying multi-config is easier than one. Maybe clash doesn't need a DB. A cache is enough.

`ReCreateSocks` is designed to create/update the socks5 server instance. if you put Authenticator in `ReCreateSocks`, you must recreate a socks5 instance according to your implementation. So split Authenticator and ReCreateSocks make hot update easier.

popu125:
> > so I thought save login informations in a file outside of all-in-one config file is easier to modify by automatic scripts.
> 
> Unless you are using the shell script, I don't think modifying multi-config is easier than one. Maybe clash doesn't need a DB. A cache is enough.
> 
> `ReCreateSocks` is designed to create/update the socks5 server instance. if you put Authenticator in `ReCreateSocks`, you must recreate a socks5 instance according to your implementation. So split Authenticator and ReCreateSocks make hot update easier.

I extracted them into `component/auth`. Is it ok?

component/auth/auth.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"sync"
+
+	C "github.com/Dreamacro/clash/constant"
+	"github.com/Dreamacro/clash/log"
+)
+
+var (
+	authenticator C.Authenticator
```
Dreamacro: `/component` and `/common` should not have a singleton, the variable should keep in where it used.

component/auth/auth.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func Authenticator() C.Authenticator {
+	return authenticator
+}
+
+func SetAuthenticator(au C.Authenticator) {
```
Dreamacro: so it's unnecessary

component/auth/auth.go
```diff
@@ -0,0 +1,54 @@
...
+	authenticator = au
+}
+
+type inMemoryAuthenticator struct {
+	sync.Map
+	logins []C.AuthUser
```
Dreamacro: emmmmm, `logins` is a good name?

component/auth/auth.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (au inMemoryAuthenticator) Verify(user string, pass string) bool {
```
Dreamacro: I think `func (au *inMemoryAuthenticator) Verify` is better

component/auth/auth.go
```diff
@@ -0,0 +1,54 @@
...
+func (au noAuthenticator) Logins() []C.AuthUser                 { return []C.AuthUser{} }
+
+func NewAuthenticator(users []C.AuthUser) C.Authenticator {
+	if len(users) == 0 {
+		return noAuthenticator{}
+	}
+
+	au := &inMemoryAuthenticator{}
+	for _, user := range users {
+		log.Infoln("Loaded user %s:%s", user.User, user.Pass)
```
Dreamacro: `/component` and `/common` should not have log

component/auth/auth.go
```diff
@@ -0,0 +1,54 @@
...
+func Authenticator() C.Authenticator {
+	return authenticator
+}
+
+func SetAuthenticator(au C.Authenticator) {
+	authenticator = au
+}
+
+type inMemoryAuthenticator struct {
+	sync.Map
```
Dreamacro: I think "extend" a sync.Map is not a good idea because it can Store and Delete by upstream

component/socks5/socks5.go
```diff
@@ -68,8 +74,63 @@ func ServerHandshake(rw io.ReadWriter) (addr Addr, command Command, err error) {
 		return
 	}
 	// write VER METHOD
-	if _, err = rw.Write([]byte{5, 0}); err != nil {
-		return
+	if authenticator.Enabled() {
```
Dreamacro: `if authenticator != nil`

component/socks5/socks5.go
```diff
@@ -68,8 +74,63 @@ func ServerHandshake(rw io.ReadWriter) (addr Addr, command Command, err error) {
...
+		if _, err = rw.Write([]byte{5, 2}); err != nil {
+			return
+		}
+
+		// Get header
+		header := []byte{0, 0}
+		if _, err = io.ReadAtLeast(rw, header, 2); err != nil {
```
Dreamacro: should be `io.ReadFull`, the following codes are the same

component/socks5/socks5.go
```diff
@@ -68,8 +74,63 @@ func ServerHandshake(rw io.ReadWriter) (addr Addr, command Command, err error) {
...
 	// write VER METHOD
-	if _, err = rw.Write([]byte{5, 0}); err != nil {
-		return
+	if authenticator.Enabled() {
+		if _, err = rw.Write([]byte{5, 2}); err != nil {
+			return
+		}
+
+		// Get header
+		header := []byte{0, 0}
```
Dreamacro: `header := make([]byte, 2)`

component/socks5/socks5.go
```diff
@@ -68,8 +74,63 @@ func ServerHandshake(rw io.ReadWriter) (addr Addr, command Command, err error) {
...
+		if _, err = io.ReadAtLeast(rw, header, 2); err != nil {
+			return
+		}
+
+		// Get username
+		userLen := int(header[1])
+		if userLen <= 0 || userLen >= MaxAuthLen {
```
Dreamacro: `[]byte` is a uint8 alias, so check userLen not equal 0 is enough

component/socks5/socks5.go
```diff
@@ -68,8 +74,63 @@ func ServerHandshake(rw io.ReadWriter) (addr Addr, command Command, err error) {
...
+			rw.Write([]byte{1, 1})
+			err = ErrGeneralFailure
+			return
+		}
+		user := make([]byte, userLen)
```
Dreamacro: you can make([]byte, 255) and reuse it

component/socks5/socks5.go
```diff
@@ -68,8 +74,63 @@ func ServerHandshake(rw io.ReadWriter) (addr Addr, command Command, err error) {
...
+		ok := authenticator.Verify(string(user), string(pass))
+		if !ok {
+			log.Infoln("Auth failed from %s", rw.RemoteAddr().String())
+			rw.Write([]byte{1, 1})
+			err = ErrGeneralFailure
+			return
+		}
+
+		// Response auth state
+		_, err = rw.Write([]byte{1, 0})
```
Dreamacro:
```go
if _, err = rw.Write([]byte{1, 0}); err != nil {
    return
}
```

constant/auth.go
```diff
@@ -0,0 +1,12 @@
+package constant
```
Dreamacro: should move to `/component`

proxy/http/server.go
```diff
@@ -29,6 +35,7 @@ func NewHttpProxy(addr string) (*HttpListener, error) {
 
 	go func() {
 		log.Infoln("HTTP proxy listening at: %s", addr)
+		authCache := cache.New(30 * time.Second)
```
Dreamacro: maybe cache should put in HttpListener

proxy/http/server.go
```diff
@@ -53,14 +60,49 @@ func (l *HttpListener) Address() string {
 	return l.address
 }
 
-func handleConn(conn net.Conn) {
+func handleAuth(conn net.Conn) {
```
Dreamacro: why put http response in a single function

component/auth/auth.go
popu125: Emmmmmm, maybe `auths` or `users`? But I think them cannot explain the true meaning too.

component/auth/auth.go
popu125: I think so. That's what my IDE done. XD

component/auth/auth.go
popu125: OK, I'll mask it.

component/socks5/socks5.go
popu125: I think it's more easy to control and prevent malformed packets by reading the packet step by step using length provided.

Dreamacro: https://tools.ietf.org/html/rfc1929

user and pass length is 1 byte, which means int(a byte) between 0-255.

```
maxAuthLen := 255
authBuf := make([]byte, maxAuthLen)

io.ReadFull(conn, authBuf[:userLen])
user := string(authBuf[:userLen]))

io.ReadFull(conn, authBuf[:passLen])
pass := string(authBuf[:passLen]))
```

it's safe.

component/auth/auth.go
Dreamacro: what about `authentication` ?
```yaml
authentication:
  - "a"
  - "b"
```

popu125: I renamed it to `local-users` in config, and `users` in code.
But still a little uncomfortable. :thinking:

popu125: > what about `authentication` ?
> 
> ```yaml
> authentication:
>   - "a"
>   - "b"
> ```

It's too long, maybe. XD

hub/executor/executor.go
```diff
@@ -36,6 +37,7 @@ func GetGeneral() *config.General {
 		Port:      ports.Port,
 		SocksPort: ports.SocksPort,
 		RedirPort: ports.RedirPort,
+		Users:     P.Authenticator().Users(),
```
Dreamacro: There is no need to expose this due to security.

hub/executor/executor.go
```diff
@@ -90,6 +92,15 @@ func updateGeneral(general *config.General) {
 	allowLan := general.AllowLan
 
 	P.SetAllowLan(allowLan)
+
+	authenticator := auth.NewAuthenticator(general.Users)
+	P.SetAuthenticator(authenticator)
+	if authenticator != nil {
```
Dreamacro: maybe should not log right now

proxy/listener.go
```diff
@@ -53,7 +64,7 @@ func ReCreateHTTP(port int) error {
 	}
 
 	var err error
-	httpListener, err = http.NewHttpProxy(addr)
+	httpListener, err = http.NewHttpProxy(addr, authenticator)
```
Dreamacro: we can read authenticator directly, so we don't need passthrough instance

popu125: authenticator is now saved in `proxy`, import it in `http` or `socks` will cause a import cycle. But I can't find a better choice to place it.

hub/executor/executor.go
popu125: maybe in `proxy` while calling `SetAuthenticator`?
Or i can't find a more suitable place.

component/auth/auth.go
```diff
@@ -0,0 +1,41 @@
...
+	if realPass, ok := au.storage.Load(user); ok && realPass == pass {
+		return true
+	}
+	return false
+}
+
+func (au *inMemoryAuthenticator) Users() []AuthUser { return au.logins }
+
+func NewAuthenticator(users []AuthUser) Authenticator {
+	if len(users) == 0 {
```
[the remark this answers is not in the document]
popu125: users length is the flag from config of enable or not auth in this scene

component/auth/auth.go
```diff
@@ -0,0 +1,41 @@
...
+	Users() []AuthUser
+}
+
+type AuthUser struct {
+	User string
+	Pass string
+}
+
+type inMemoryAuthenticator struct {
+	storage sync.Map
```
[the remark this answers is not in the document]
popu125: :sweat_smile:

component/auth/auth.go
```diff
@@ -0,0 +1,41 @@
...
+	logins  []AuthUser
+}
+
+func (au *inMemoryAuthenticator) Verify(user string, pass string) bool {
```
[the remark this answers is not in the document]
popu125: Yeah, it's more clear.

hub/executor/executor.go
popu125: I thought there should be some users need update auth info by web api. Or should I rewrite and return only username?

component/auth/auth.go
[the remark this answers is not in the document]
Dreamacro: pass through an empty []AuthUser and it should never return true when called Verify

hub/executor/executor.go
Dreamacro: return the only username is enough

component/auth/auth.go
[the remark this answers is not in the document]
Dreamacro: `storage *sync.Map`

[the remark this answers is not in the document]
popu125: I know what you mean. Just a little embarrassed.

component/auth/auth.go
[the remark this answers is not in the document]
popu125: But sometimes user does not need an auth. They keep the config blank and the program knows they want to disable auth.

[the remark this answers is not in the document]
Dreamacro: https://play.golang.org/p/Y_9xz_IbEHV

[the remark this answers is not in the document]
popu125: But it's nil when you return an interface. https://play.golang.org/p/4doPP7IoP74

proxy/http/server.go
```diff
@@ -53,14 +60,45 @@ func (l *HttpListener) Address() string {
 	return l.address
 }
 
-func handleConn(conn net.Conn) {
+func doAuth(loginStr string, auth auth.Authenticator, cache *cache.Cache) (ret bool) {
```
Dreamacro: I prefer called `canActivate`, and what about

```go
func canActivate(loginStr string, auth auth.Authenticator, cache *cache.Cache) bool {
	if result := cache.Get(loginStr); result != nil {
		return result.(bool)
	}

	loginData, err := base64.StdEncoding.DecodeString(loginStr)
	login := strings.SplitN(string(loginData), ":", 2)
	ret := err == nil && len(login) == 2 && auth.Verify(login[0], login[1])
 	cache.Put(loginStr, ret, time.Minute)
	return ret
}
```

proxy/http/server.go
```diff
@@ -53,14 +60,45 @@ func (l *HttpListener) Address() string {
...
 	if err != nil || request.URL.Host == "" {
 		conn.Close()
 		return
 	}
 
+	if auth != nil {
+		if authStrings := strings.Split(request.Header.Get("Proxy-Authorization"), " "); len(authStrings) != 2 {
+			_, err = conn.Write([]byte("HTTP/1.1 407 Proxy Authentication Required\r\nProxy-Authenticate: Basic\r\n\r\n"))
+			return
+		} else {
```
Dreamacro: Unnecessary else

Dreamacro: or just else if

component/auth/auth.go
Dreamacro: I still recommend `authentication` if you don't have another variable name

popu125: OK, you convinced me.

proxy/listener.go
Dreamacro: you can create `/proxy/auth` to store it

popu125: It looks a little strange to create an `auth` besides three proxy methods. But it's ok if you're permitted.

hub/executor/executor.go
```diff
@@ -109,8 +110,10 @@ func updateGeneral(general *config.General) {
 
 func updateUsers(users []auth.AuthUser) {
 	authenticator := auth.NewAuthenticator(users)
-	P.SetAuthenticator(authenticator)
-	for _, username := range authenticator.Users() {
-		log.Infoln("Loaded user: %s", username)
+	authStore.SetAuthenticator(authenticator)
+	if authenticator != nil{
```
Dreamacro: should not log each username just log `Loaded new Authenticator`

popu125: I has thought it's more convenient to detect if some of users config wrong or not saved by logging every username.

Dreamacro: it's not good for stdout log, there would have an `event` api in the future to catch all of clash event

hub/executor/executor.go
```diff
@@ -111,9 +111,7 @@ func updateGeneral(general *config.General) {
 func updateUsers(users []auth.AuthUser) {
 	authenticator := auth.NewAuthenticator(users)
 	authStore.SetAuthenticator(authenticator)
-	if authenticator != nil{
-		for _, username := range authenticator.Users() {
-			log.Infoln("Loaded user: %s", username)
-		}
+	if authenticator != nil {
+		log.Infoln("Authentication of local server is enabled")
```
Dreamacro: it's so weird for hot update

Dreamacro: this is final modify

README.md
```diff
@@ -109,6 +109,11 @@ external-controller: 127.0.0.1:9090
 experimental:
   ignore-resolve-fail: true # ignore dns resolve fail, default value is true
 
+# authenticate of local SOCKS5/HTTP(S) server
```
Dreamacro: `authenticate` --> `authentication` typo

proxy/http/server.go
```diff
@@ -53,14 +61,39 @@ func (l *HttpListener) Address() string {
...
 	if err != nil || request.URL.Host == "" {
 		conn.Close()
 		return
 	}
 
+	authenticator := authStore.Authenticator()
+	if authenticator != nil {
+		if authStrings := strings.Split(request.Header.Get("Proxy-Authorization"), " "); len(authStrings) != 2 {
+			_, err = conn.Write([]byte("HTTP/1.1 407 Proxy Authentication Required\r\nProxy-Authenticate: Basic\r\n\r\n"))
+			return
```
Dreamacro: missing `conn.Close()`

proxy/http/server.go
```diff
@@ -53,14 +61,39 @@ func (l *HttpListener) Address() string {
...
 }
 
-func handleConn(conn net.Conn) {
+func canActivate(loginStr string, authenticator auth.Authenticator, cache *cache.Cache) (ret bool) {
+	if result := cache.Get(loginStr); result != nil {
+		ret = result.(bool)
+	}
+	loginData, err := base64.StdEncoding.DecodeString(loginStr)
+	login := strings.Split(string(loginData), ":")
+	ret = err == nil && len(login) != 2 && authenticator.Verify(login[0], login[1])
```
Dreamacro: `login == 2`

Dreamacro:
Thanks for your contribution and patience

Fndroid:
https://github.com/Dreamacro/clash/blob/e5284cf647717a8087a185d88d15a01096274bc2/hub/executor/executor.go#L26-L30

端口都在General里，为什么auth不放在General里或者跟随General一起在force下才更新呢？

abae136ab:
> https://github.com/Dreamacro/clash/blob/e5284cf647717a8087a185d88d15a01096274bc2/hub/executor/executor.go#L26-L30
> 
> 端口都在General里，为什么auth不放在General里或者跟随General一起在force下才更新呢？
我正想来问这个问题，发现F大已经帮我问了，现在每次更新订阅都要加一下
#authentication of local SOCKS5/HTTP(S) server
比较麻烦，建议改到general YML配置文件内，不要随着profile走，这样就一劳永逸了。

Dreamacro:
@Fndroid @abae136ab 对于这个疑问, 你们有没有发现目前 force 影响的是 root config 上的非复杂单层配置, 至于 force 到底应该影响哪部分配置, 可以单独开一个 issue 来讨论, 再决定一个最符合这个字段的行为

PR #253, opened 2019-08-07
fix the yaml/yml compatibility issue

ruisiji:
fixes https://github.com/Dreamacro/clash/issues/235

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -106,7 +106,11 @@ func readRawConfig(path string) ([]byte, error) {
 	}
 
 	path = path[:len(path)-5] + ".yml"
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -106,7 +106,12 @@ func readRawConfig(path string) ([]byte, error) {
 	}
 
 	path = path[:len(path)-5] + ".yml"
-	return ioutil.ReadFile(path)
+	_, err = os.Stat(path)
```
Dreamacro:
```go
if _, err = os.Stat(path); err == nil {
```

Issue #135, opened 2019-03-12
port和socks-port能否指定监听的IP地址


zhaibin18:
目前配置文件这样写会报错
port: 192.168.1.1:7890

Dreamacro:
有什么特殊的场景吗

zhaibin18:
路由器上，不想监听ppp0，只监听内部给局域网内用，allow-lan: true会监听所有接口。

comzyh:
iptables 或者 firewalld 过滤下？

zhaibin18 <notifications@github.com> 于 2019年3月16日周六 19:34写道：

ruisiji:
@Dreamacro I'd like to replace `allowLan` with `bindAddress` to allow fine-grain control the bind address.

This is very helpful when a device has multiple interfaces, especially has a public IP address.

If you agree this approach, I can make a PR to implement it.

Dreamacro:
@ruisiji replace `allowLan` is a big break change, both GUI and dashboard will be affected. I don't recommend to do it.

ruisiji:
We can keep the backward compatibility, something like allow setting `bindAddress` to `true/false/x.x.x.x`, `true` is equal to `0.0.0.0` and `false` equal to `127.0.0.1`.

Dreamacro:
```bindAddress: true``` look so weird. maybe we can provide a new config like `lan-address` to set specific address if `allow-lan` is true.

**Comment by ruisiji** [comment]
I don't know how does GUI and dashboard code work, but the logic should be easy to implement when switched to `bindAddress`.

* `allowLan: true` ==> `bindAddress: 0.0.0.0`
* `allowLan: false` ==> `bindAddress: 127.0.0.1`

We can provide enough time for GUI and dashboard to change the code logic before we deprecated the `allowLan`.

Introducing another extra parameter is a little complex for users since the use case can be done with a single parameter.

**Comment by Dreamacro** [comment]
But in most cases (even for me), don't need to bind a specific address. For most users is enough now.

```yaml
port: 7890
allow-lan: true # false
```

for advanced users, just add an external config

```yaml
port: 7890
allow-lan: true # false
lan-address: 192.168.0.2 # ==> 192.168.0.2:7890 for HTTP
```
the `lan-address` default is `:` , because `:` would bind dual stack address (:: and 0.0.0.0)

ruisiji:
So the final approach will be something like:

[...]

PR #255, opened 2019-08-07
Feature: support customizing bind-address when allow-lan is true

ruisiji:
fixes https://github.com/Dreamacro/clash/issues/135

### DIFF
--- a/proxy/listener.go
+++ b/proxy/listener.go
@@ -11,7 +11,8 @@ import (
 )
 
 var (
-	allowLan = false
+	allowLan    = false
+	bindAddress = "*"
 
 	socksListener    *socks.SockListener
 	socksUDPListener *socks.SockUDPListener
@@ -34,12 +35,20 @@ func AllowLan() bool {
 	return allowLan
 }
 
+func BindAddress() string {
+	return bindAddress
+}
+
 func SetAllowLan(al bool) {
 	allowLan = al
 }
 
+func SetBindAddress(host string) {
+	bindAddress = host
+}
+
 func ReCreateHTTP(port int) error {
-	addr := genAddr(port, allowLan)
+	addr := genAddr(bindAddress, port, allowLan)
 
 	if httpListener != nil {
 		if httpListener.Address() == addr {
@@ -63,7 +72,7 @@ func ReCreateHTTP(port int) error {
 }
 
 func ReCreateSocks(port int) error {
-	addr := genAddr(port, allowLan)
+	addr := genAddr(bindAddress, port, allowLan)
 
 	if socksListener != nil {
 		if socksListener.Address() == addr {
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -11,6 +11,7 @@ env:
     - GO111MODULE=on
 script:
   - go test ./...
+  - make linux-amd64
```
Dreamacro: Why is it built?

constant/general.go
```diff
@@ -0,0 +1,4 @@
+package constant
```
Dreamacro: Variables cannot be defined here

proxy/listener.go
```diff
@@ -5,13 +5,15 @@ import (
...
 
+	C "github.com/Dreamacro/clash/constant"
 	"github.com/Dreamacro/clash/proxy/http"
 	"github.com/Dreamacro/clash/proxy/redir"
 	"github.com/Dreamacro/clash/proxy/socks"
 )
 
 var (
-	allowLan = false
+	allowLan    = C.AllowLan
```
Dreamacro: use `clash/constant` value let the data flow becomes chaotic

proxy/listener.go
```diff
@@ -97,10 +105,6 @@ func reCreateSocksUDP(port int) error {
 		socksUDPListener = nil
 	}
 
-	if portIsZero(addr) {
```
Dreamacro: why remove this check function

.travis.yml
ruisiji: this can make the ci more reliable.

proxy/listener.go
ruisiji: because this check is done in the caller function, no need here.

constant/general.go
ruisiji: why can not use constant for these two variables? could you give me more info?

proxy/listener.go
ruisiji: could you provide more info about this?

.travis.yml
Dreamacro: Is there anything that `go test` cannot do and `make` can do?

constant/general.go
Dreamacro: `clash/constant` is used to store abstract instead of the default config value

proxy/listener.go
Dreamacro: `allowLan` in `clash/proxy` just a private variable, It doesn't really matter what value it initially has.

proxy/listener.go
Dreamacro: I don't see where to validate this value

ruisiji: caller function: https://github.com/Dreamacro/clash/blob/288afd1308b95f61585233dd5bdc9bbebe8f96de/proxy/listener.go#L65

validation: https://github.com/Dreamacro/clash/blob/288afd1308b95f61585233dd5bdc9bbebe8f96de/proxy/listener.go#L76

call function: https://github.com/Dreamacro/clash/blob/288afd1308b95f61585233dd5bdc9bbebe8f96de/proxy/listener.go#L86

validation in call function(no need): https://github.com/Dreamacro/clash/blob/288afd1308b95f61585233dd5bdc9bbebe8f96de/proxy/listener.go#L100

Dreamacro:
thanks for your contribution 👏

PR #276, opened 2019-08-22
Feature: support speed test of URLTest through API

tslling:
support manually triggering a speed test of URLTest proxy group through API. #273


hub/route/proxies.go
```diff
@@ -135,3 +136,14 @@ func getProxyDelay(w http.ResponseWriter, r *http.Request) {
...
+
+func reelectProxy(w http.ResponseWriter, r *http.Request) {
+	proxy := r.Context().Value(CtxKeyProxy).(*A.Proxy)
+	urlTest, ok := proxy.ProxyAdapter.(*A.URLTest)
+	if !ok {
+		render.Status(r, http.StatusBadRequest)
+		render.JSON(w, r, newError("Must be a Url-Test"))
+		return
+	}
+	urlTest.SpeedTest()
```
popu125: Maybe it need a response.

adapters/outbound/urltest.go
```diff
@@ -71,6 +71,10 @@ func (u *URLTest) Destroy() {
 	u.done <- struct{}{}
 }
 
+func (u *URLTest) SpeedTest() {
+	u.speedTest()
```
popu125: I think its better to do speedtest concurrently.

The implement of `speedTest()` does prevent race condition, it's safe to just add a `go`.

popu125:
个人对 API 名字的选择持保留态度（虽然并没有什么卵用

hub/route/proxies.go
tslling: You mean response before speedtest? By doing this, we return nothing useful to client, how about waiting until the speedtest finished, then we can return the picked proxy to client. But obviously we can not wait all the time, maybe client will supply a ```timeout```.

PR #280, opened 2019-08-25
Feature: add `prefer` option to url-test

comwrg:
two options
* fast
* stable

In `fast` mode, fastest server may disconnect cause embarrassment (https://github.com/Dreamacro/clash/issues/273)

`stable` mode may avoid this case


comwrg:
Resolved in https://github.com/Dreamacro/clash/pull/283.

PR #282, opened 2019-08-27
optimize

comwrg:
If all proxies complete speed test within `defaultURLTestTimeout` seconds, `<-ctx.Done()` will lead to block, goroutine can be released until `defaultURLTestTimeout` seconds later.

### DIFF
--- a/adapters/outbound/urltest.go
+++ b/adapters/outbound/urltest.go
@@ -125,8 +125,6 @@ func (u *URLTest) speedTest() {
 	if fast != nil {
 		u.fast = fast.(C.Proxy)
 	}
-
-	<-ctx.Done()
 }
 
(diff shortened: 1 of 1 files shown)


Dreamacro:
It is acceptable to block `speedTest` for 5 seconds. If not blocked, will lead to competition.

comwrg:
https://github.com/Dreamacro/clash/blob/71f0a4e35ea01b332387ac5ee78e900abd0abb04/adapters/outbound/urltest.go#L106-L109
The code can keep only one goroutine in this function at same time I think.

Dreamacro:
The `atomic code` is used to reduce the call of speedTest. But it‘s indeed a dead code.

Because of the max running time of `picker.Wait()` still 5s.

PR #283, opened 2019-08-28
Optimize url-test dial

comwrg:


Dreamacro:
Does this change make url-test more complicated?

When the proxy node is stable, we don't need fallback.

When the proxy node is unstable, reduce the interval is effective. speedTest plays the role of the health check.

comwrg:
Fix (https://github.com/Dreamacro/clash/pull/283/commits/0348ab18f57c79e095155f7a158cbc77a28b327a, https://github.com/Dreamacro/clash/pull/283/commits/ea72a3faad5e8e112783b362b8bafe7d878d0a4d)
If current proxy node is unavailable, function `fallback` will be executed many times at same time. Only once enough at same time.

Optimize (https://github.com/Dreamacro/clash/pull/283/commits/a1960171377e5c45e958f376b2d6d8068d3fe48a)
If current proxy node is unavailable, attempting to send connections that could not be sent before to the new proxy. 
Optimize experience. There is no need to constantly refresh the webpage when current proxy node is unavailable.

Dreamacro:
The difference between `fallback` and `speedTest` is that `speedTest` are some http requests, it's expensive, but `fallback` just a for-loop.

> If current proxy node is unavailable, attempting to send connections that could not be sent before to the new proxy.
> Optimize experience. There is no need to constantly refresh the webpage when current proxy node is unavailable.

reduce the interval is effective.

comwrg:
But when current proxy node is unavailable will lead to browser display `the website cannot be reached`, then we need to refresh the webpage constantly, it's so terrible.

Dreamacro:
When you use a lower interval value, the current node should be completely effective. If you see `the website cannot be reached` frequently, I think you would like to replace your vps. if you really need fallback dial, add `fast.Dial` after `fallback()` is enough.

comwrg:
OK. But It is also necessary to lock function `fallback`, if `fallback` is process, this mean that current proxy is unavailable, so all connections need wait for `fallback` finished to choice a valid proxy node.

Dreamacro:
`fallback` cost lower and the number of concurrencies will not large than 10. I still don't recommend doing other things.

BTW, https://godoc.org/golang.org/x/sync/singleflight maybe is what you want. we don't need another `atomic`.

PR #288, opened 2019-08-31
Fix: HTTP Status Code `100 Continue` support"

soarqin:
https://tools.ietf.org/html/rfc7231#section-6.2.1

This is just a temporary fix for it:
1. remove `Expect: 100-continue` from request header
2. if got `100 Continue` by accident, just read another request

To fix it completely, we must(but no need unless you care about the wait time of 100-contine):
1. read only headers and check `Expect: 100-continue`, if not found, just continue to read body and send the whole request and skip all steps followed
2. send headers without data body
3. read response with status code `100 Continue`, if timeout or got incorrect status code, close the connection with certain status code returned (skip step 4)
4. read data body, send it and read response

### DIFF
--- a/tunnel/connection.go
+++ b/tunnel/connection.go
@@ -16,48 +16,49 @@ func (t *Tunnel) handleHTTP(request *adapters.HTTPAdapter, outbound net.Conn) {
 	conn := newTrafficTrack(outbound, t.traffic)
 	req := request.R
 	host := req.Host
-	brconn := bufio.NewReader(conn)
-	brreq := bufio.NewReader(request)
+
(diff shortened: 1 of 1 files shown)


Dreamacro:
Is there any real-world application use `100 Continue` for request?

soarqin:
In Mono/.NET Program, System.Net.Http.WebClient sends `Expect: 100-continue` on POST, event with small data.
That's why I got stuck and tried to resolve the problem by http packets capture and analyze.

Dreamacro:
I adjust the style of code, Is this fit your expectation?

soarqin:
yeah, good job.

PR #294, opened 2019-09-09
Fix: don't close connection on status `100 Continue` and header `Proxy-Connection: Keep-Alive`

soarqin:
1. if `Connection: Keep-Alive` is set in header, do not break loop. Otherwise http client side may think the connection was keep-alive but reset by remote http server
2. if client sends keep-alive, follow its request even there is no data reply

this is just another bug found in my .NET/Mono WebClient program

### DIFF
--- a/tunnel/connection.go
+++ b/tunnel/connection.go
@@ -37,7 +37,16 @@ func (t *Tunnel) handleHTTP(request *adapters.HTTPAdapter, outbound net.Conn) {
 			break
 		}
 		adapters.RemoveHopByHopHeaders(resp.Header)
-		if resp.ContentLength >= 0 {
+
+		if resp.StatusCode == http.StatusContinue {
+			err = resp.Write(request)
+			if err != nil {
+				break
+			}
+			goto handleResponse
+		}
+
+		if keepAlive || resp.ContentLength >= 0 {
 			resp.Header.Set("Proxy-Connection", "keep-alive")
 			resp.Header.Set("Connection", "keep-alive")
 			resp.Header.Set("Keep-Alive", "timeout=4")
@@ -48,14 +57,6 @@ func (t *Tunnel) handleHTTP(request *adapters.HTTPAdapter, outbound net.Conn) {
 		err = resp.Write(request)
 		if err != nil || resp.Close {
 			break
-		}
-
-		if !keepAlive {
-			break
-		}
-
-		if resp.StatusCode == http.StatusContinue {
-			goto handleResponse
 		}
 
 		req, err = http.ReadRequest(inboundReeder)


Dreamacro:
Can you provide the HTTP request message about unexpected return?

soarqin:
just send a http request without `Connection: Keep-Alive` (would be translated to `Proxy-Connection: Keep-Alive` while using http proxy). If there is data in response, a `Connection: Keep-Alive` would be set at https://github.com/Dreamacro/clash/blob/dev/tunnel/connection.go#L41, but the loop is break out at https://github.com/Dreamacro/clash/blob/dev/tunnel/connection.go#L54 so that the connection would be closed by server side, which does not satifsty the `Keep-Alive` logic. So clients without auto-reconnection mechanisim would get an unexpected connection reset here.

Dreamacro:
But your change seems that the `keepAlive` become useless

soarqin:
I removed `keepAlive` break check just beacause it is always opposite to `resp.Close`, but read the changes carefully and you should understand that it is not useless.

Dreamacro:
@soarqin I think this may better than your code change

```diff
for {
	keepAlive := strings.TrimSpace(strings.ToLower(req.Header.Get("Proxy-Connection"))) == "keep-alive"
+       shouldWaitAnother := true

	req.Header.Set("Connection", "close")
	req.RequestURI = ""
	adapters.RemoveHopByHopHeaders(req.Header)
	err := req.Write(conn)
	if err != nil {
		break
	}

handleResponse:
	resp, err := http.ReadResponse(outboundReeder, req)
	if err != nil {
		break
	}
	adapters.RemoveHopByHopHeaders(resp.Header)
	if resp.ContentLength >= 0 {
		resp.Header.Set("Proxy-Connection", "keep-alive")
		resp.Header.Set("Connection", "keep-alive")
		resp.Header.Set("Keep-Alive", "timeout=4")
		resp.Close = false
	} else {
+               shouldWaitAnother = false
		resp.Close = true
	}
	err = resp.Write(request)
	if err != nil || resp.Close {
		break
	}

-       if !keepAlive {
+	if !shouldWaitAnother && !keepAlive {
		break
	}

	if resp.StatusCode == http.StatusContinue {
		goto handleResponse
	}

	req, err = http.ReadRequest(inboundReeder)
	if err != nil {
		break
	}

	// Sometimes firefox just open a socket to process multiple domains in HTTP
	// The temporary solution is close connection when encountering different HOST
	if req.Host != host {
		break
	}
}
```

soarqin:
yup, your changes are working fine in my test.
but I'm thinking of `Proxy-Connection: Keep-Alive`, I've not found any doc about if keep-alive is strictly needed on receiving it, but keep it alive is not bad in my idea.

soarqin:
also I prefer heavily optimized codes than better reading codes, so that I think my codes of removing a check is better than adding a new variable and checking it on every request.
PS: swap TrimSpace and ToLower in the line `keepAlive := ` is also an optimization to reduce possible calculations.

Dreamacro:
I prefer better reading codes than heavily optimized codes when code not in the "hot path". It helps me reduce the burden of maintenance.

BTW `keepAlive = true` makes no sense because of the `keepAlive` scope is in `for` and has not been reused since then.

soarqin:
you are right, that was my mistake.
I found another problem that connection with responsed status 100 Continue should not be closed, so I fixed my codes for it and to correspond to your notes.

PS: I didn't mean bad reading codes are good. But optimizations without break graceful codes too much are better. And if tricky codes were written, certain comments are required in order to help others' read.

Dreamacro:
thanks for your contribution

PR #308, opened 2019-09-18
enable rest api tls

ruisiji:
This PR introduces two parameters:

* tls-cert-file the external controller TLS certificate file path
* tls-private-key-file the external controller TLS private key file path

Setting these two parameters will make external controller host under TLS connection.

If the file path does not exist, clash will throw error message and does not start external controller.

```
ERRO[0000] External controller error: open domain.crt: no such file or directory
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,10 @@ log-level: info
 # A RESTful API for clash
 external-controller: 127.0.0.1:9090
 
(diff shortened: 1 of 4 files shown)


Dreamacro:
you can do this over a caddy or nginx, the application usually won't provide tls in real-world

ruisiji:
ok.

PR #311, opened 2019-09-19
fix deadlock

comwrg:
When error is `no such host` will lead to deadlock.
https://github.com/Dreamacro/clash/blob/8adcc4d83b487c6ba143a5f162d26b113a8ae26f/adapters/outbound/util.go#L128-L135
https://github.com/Dreamacro/clash/blob/8adcc4d83b487c6ba143a5f162d26b113a8ae26f/adapters/outbound/util.go#L156-L173

### DIFF
--- a/adapters/outbound/util.go
+++ b/adapters/outbound/util.go
@@ -131,6 +131,7 @@ func dialTimeout(network, address string, timeout time.Duration) (net.Conn, erro
 			ip, err = dns.ResolveIPv4(host)
 		}
 		if err != nil {
+			results <- dialResult{Conn: nil, error: err, ipv6: ipv6, done: true}
 			return
 		}
 


Dreamacro:
your change may cause goroutine leak, I would push a fixed commit on this pr.

comwrg:
Looks good.

By the way, I have some doubts about this code. What is the purpose of the code?
https://github.com/comwrg/clash/blob/e8e3bf0bb9254ddd075dc21d94146d1772a6ac0f/adapters/outbound/util.go#L129-L132

Dreamacro:
In order to describe more accurately, I use Chinese.

`returned` 主要是用来做资源清理的，倘若同时 dial ipv6 ipv4 成功，由于函数只需要一个 net.Conn，所以「较慢」的那个 net.Conn 应该被关闭，你之前改动会一直阻塞在把结果传递到 `results` 当中，造成不必要的 goroutine 泄漏。这里参考了官方库的实现 https://github.com/golang/go/blob/master/src/net/dial.go#L448-L518

comwrg:
了解了。个人感觉可以写一点注释，复杂度有点高。

comzyh:
I have a question. Why you are using two timeouts?
```go
	dialer := net.Dialer{Timeout: timeout}
	ctx, cancel := context.WithTimeout(context.Background(), timeout)	
```
Maybe the second one can be replaced with `WithCancel`?

Other part looks good to me,

Dreamacro:
@comzyh remove first timeout

comzyh:
LGTM

PR #325, opened 2019-09-24
Skip dual-stack dial when ip address is provided directly

comzyh:
When an IP address is provided, we should not start the dual-stack dial procedure.

---

Sometimes, when users are connecting remote host with **IP address** through clash

```
 curl --socks5-hostname 127.0.0.1:7891 http://224.0.0.0
```
For now, Clash will output one line warning

> WARN[0000] dial DIRECT error: dial tcp6: address 224.0.0.0: no suitable address found 

Because if both `IPv4` and `IPv6` are failed to connect, `dialTimeout` will return the `IPv6` error
https://github.com/Dreamacro/clash/blob/1a8a6d0b5d28161069c23560464be9f1e0eb6fd7/adapters/outbound/util.go#L163-L167

Actually, when users are using IPv4 address, clash should not try to connect with IPv6.

### DIFF
--- a/adapters/outbound/util.go
+++ b/adapters/outbound/util.go
@@ -110,7 +110,7 @@ func dialTimeout(network, address string, timeout time.Duration) (net.Conn, erro
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
-	// if `host` is an ipv4/ipv6 address rather than hostname, we should not start dual stack race
+	// if `host` is an ipv4/ipv6 address rather than a DNS name, we should not start the dual-stack race
 	if net.ParseIP(host) != nil {
 		if strings.Contains(host, ".") {
 			return dialer.DialContext(ctx, "tcp4", address)
--- a/adapters/outbound/util.go
+++ b/adapters/outbound/util.go
@@ -8,7 +8,6 @@ import (
 	"net"
 	"net/url"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
@@ -110,15 +109,6 @@ func dialTimeout(network, address string, timeout time.Duration) (net.Conn, erro
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
-	// if `host` is an ipv4/ipv6 address rather than a DNS name, we should not start the dual-stack race
-	if net.ParseIP(host) != nil {
-		if strings.Contains(host, ".") {
-			return dialer.DialContext(ctx, "tcp4", address)
-		} else {
-			return dialer.DialContext(ctx, "tcp6", address)
-		}
-	}
-
 	returned := make(chan struct{})
 	defer close(returned)
 
@@ -170,7 +160,7 @@ func dialTimeout(network, address string, timeout time.Duration) (net.Conn, erro
 				return res.Conn, nil
 			}
 
-			if res.ipv6 {
+			if !res.ipv6 {
 				primary = res
 			} else {
 				fallback = res
--- a/adapters/outbound/util.go
+++ b/adapters/outbound/util.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/url"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


adapters/outbound/util.go
```diff
@@ -109,6 +110,15 @@ func dialTimeout(network, address string, timeout time.Duration) (net.Conn, erro
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
+	// if `host` is an ipv4/ipv6 address rather than a DNS name, we should not start the dual-stack race
+	if net.ParseIP(host) != nil {
+		if strings.Contains(host, ".") {
```
comzyh: The same logic as 
https://github.com/golang/go/blob/1804bbab6285754f69a0683e60fc5590429dc1d1/src/net/ip.go#L675-L685

Dreamacro:
https://github.com/Dreamacro/clash/blob/dev/dns/iputil.go#L20

comzyh:
> https://github.com/Dreamacro/clash/blob/dev/dns/iputil.go#L20

I have noticed that. What is your concern?

Dreamacro:
why write defensive code again?

comzyh:
> why write defensive code again?

The real problem is how to avoid it, and I have not come out with a good idea.

`ParseIP` in this PR and in `ResolveIPv4` have a different purpose. 

If the `host` is really an IPv4 address, the `ParseIP` checking  `ResolveIPv4` cannot avoid the IPv6D dialing.

adapters/outbound/util.go
```diff
@@ -109,6 +110,15 @@ func dialTimeout(network, address string, timeout time.Duration) (net.Conn, erro
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
+	// if `host` is an ipv4/ipv6 address rather than a DNS name, we should not start the dual-stack race
+	if net.ParseIP(host) != nil {
```
comzyh: If this condition is met, this function must return.

Dreamacro:
> The real problem is how to avoid it, and I have not come out with a good idea.

What do you want to avoid?

comzyh:
> > The real problem is how to avoid it, and I have not come out with a good idea.
> 
> What do you want to avoid?
Answer to this question:
As you said, there is another `ParseIP` in https://github.com/Dreamacro/clash/blob/dev/dns/iputil.go#L20
If the function parameter `address` **is a DNS name**, in this PR, `ParseIP` will be called twice, once in `if net.ParseIP(host) != nil` and once in `ResolveIPv4`/`ResolveIPv6`.

Dreamacro:
@comzyh cc

adapters/outbound/util.go
```diff
@@ -160,7 +160,7 @@ func dialTimeout(network, address string, timeout time.Duration) (net.Conn, erro
 				return res.Conn, nil
 			}
 
-			if res.ipv6 {
```
comzyh: Yes, this solves the problem when using IPv4 Address. But raise a new problem when using ipv6 address.

dns/iputil.go
```diff
@@ -49,8 +53,11 @@ func ResolveIPv6(host string) (net.IP, error) {
 	}
 
 	ip := net.ParseIP(host)
-	if ip6 := ip.To16(); ip6 != nil {
-		return ip6, nil
+	if ip != nil {
+		if ip6 := ip.To16(); ip6 != nil {
```
comzyh: Believe me, this won't work. you can have a try https://play.golang.org/p/J-jbRBe0SH1
```go
package main

import (
	"fmt"
	"net"
)

func main() {
	fmt.Println(net.ParseIP("1.2.3.4").To16())
}
```

adapters/outbound/util.go
comzyh: I have tried your code c1dab973e1a28468bd1e55c22cf0691927f0ddc6

```bash
curl --socks5-hostname 127.0.0.1:1081 http://\[2001:db8::1\] 
```
As I said, it won't return the correct error message.

> time="2019-09-26T01:26:55+08:00" level=debug msg="[DNS] resolve 2001:db8::1 error: All DNS requests failed"
> time="2019-09-26T01:27:32+08:00" level=warning msg="dial DIRECT error: ip version error"

Dreamacro:
@comzyh I fixed some ip version check and throw the error correctly

comzyh:
LGTM.
Thanks

PR #326, opened 2019-09-26
optimize time takes to get fastest node

comwrg:
Optimize time takes to get fastest node especial when startup and the first node is not alive.

### DIFF
--- a/adapters/outbound/urltest.go
+++ b/adapters/outbound/urltest.go
@@ -123,10 +123,12 @@ func (u *URLTest) speedTest() {
 		})
 	}
 
-	fast := picker.Wait()
+	fast := picker.WaitFirstResult()
 	if fast != nil {
 		u.fast = fast.(C.Proxy)
 	}
+
+	picker.Wait()
 }
 
 func NewURLTest(option URLTestOption, proxies []C.Proxy) (*URLTest, error) {
--- a/adapters/outbound/urltest.go
+++ b/adapters/outbound/urltest.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"net"
-	"sync"
 	"sync/atomic"
 	"time"
 
@@ -113,7 +112,6 @@ func (u *URLTest) speedTest() {
 
 	picker, ctx, cancel := picker.WithTimeout(context.Background(), defaultURLTestTimeout)
 	defer cancel()
-	var once sync.Once
 	for _, p := range u.proxies {
 		proxy := p
 		picker.Go(func() (interface{}, error) {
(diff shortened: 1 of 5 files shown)


Dreamacro:
https://github.com/Dreamacro/clash/blob/50b4bc8faeacf29405a30320cc47871fa549c49d/common/picker/picker.go#L47-L62

comwrg:
In the original code, the program want to get the fastest node need to wait for all nodes return, it may needs to wait for 5 seconds in the worst case.
And now the program only need wait for fastest nodes return. We can use the fastest node immediately without waiting for all nodes to return.

For example, when the program is started, the program will call method `speedTest`, if the first node is unavailable coincidentally (the first node is deafult node if fastest node not out https://github.com/Dreamacro/clash/blob/34338e7107c1868124f8aab2446f6b71c9b0640f/adapters/outbound/urltest.go#L155), we may need wait for a few seconds if we want to visit web page.

Dreamacro:
the picker return immediately, PoC

```go
func TestPicker_Immediately(t *testing.T) {
	picker, ctx := WithContext(context.Background())
	picker.Go(sleepAndSend(ctx, 300, 2))
	picker.Go(sleepAndSend(ctx, 20, 1))

	start := time.Now()

	number := picker.Wait()
	if number != nil && number.(int) != 1 {
		t.Error("should recv 1", number)
	}

	if time.Since(start) > time.Millisecond*30 {
		t.Error("should return immediately", number)
	}
}
```

comwrg:
https://github.com/Dreamacro/clash/blob/904c354ee4c45a25849e7b821f0ae2ff83cf76a0/adapters/outbound/urltest.go#L113
The method `speedTest` use `WithTimeout` not `WithContext`, and the comment show
https://github.com/Dreamacro/clash/blob/904c354ee4c45a25849e7b821f0ae2ff83cf76a0/common/picker/picker.go#L29



```go
func TestPicker_Immediately(t *testing.T) {
	picker, ctx, cancel := WithTimeout(context.Background(), time.Millisecond*500)
	defer cancel()
	picker.Go(sleepAndSend(ctx, 300, 2))
	picker.Go(sleepAndSend(ctx, 20, 1))

	start := time.Now()

	number := picker.Wait()
	if number != nil && number.(int) != 1 {
		t.Error("should recv 1", number)
	}

	if time.Since(start) > time.Millisecond*30 {
		t.Error("should return immediately", number)
	}
}
```

Dreamacro:
maybe you should modify `picker` insteadof add side effect in `picker.Go`

tslling:
while this makes the picker return quickly, it does not really cancel other tasks.
https://github.com/Dreamacro/clash/blob/904c354ee4c45a25849e7b821f0ae2ff83cf76a0/common/picker/picker.go#L29
why not cancel those tasks.

comwrg:
> while this makes the picker return quickly, it does not really cancel other tasks.
> https://github.com/Dreamacro/clash/blob/904c354ee4c45a25849e7b821f0ae2ff83cf76a0/common/picker/picker.go#L29
> 
> 
> why not cancel those tasks.

Need to record delay of each node.
https://github.com/Dreamacro/clash/blob/d3c50cf89fb74f02e91211226009a2cd924a72ec/adapters/outbound/base.go#L142-L152

tslling:
> Need to record delay of each node.

ok, I got it. But it seems only this specific call needs this feature, not all of them.

on the other hand,  ```speedTest``` is only responsible for "find the fastest proxy", getting delay of other proxies looks like a side effect, since we already have a separete API to do this job.

Dreamacro:
@comwrg cc

comwrg:
LGTM

PR #341, opened 2019-10-04
add interface for gomobile bind

oasiscifr:
So we can embed clash inside other Proxy APPs which do not have the power of rules.

To use clash as an android library (aar file) in Android service thread, I have to add context to each server, so that I can stop clash gracefully rather than kill it (the APP will quit at the same time).

 :)


Dreamacro:
I think the best way for the third party application is that write a go file in their repository. (like https://github.com/yichengchen/clashX/blob/master/ClashX/main.go). clash repo can modify or provide the low-level API.

oasiscifr:
@Dreamacro Thank you for your reply!
I think writing a go file in dependent repos is okay.
But I've met a problem, a method to stop all listeners gracefully in thread, rather than killing, is required. Sending a SIGINT massage would terminate the main thread of Android APP.
Can I achieve this without modifying code of Clash?

Dreamacro:
There is a hacker way to stop all listener, send `0 port` to API:

```
$ curl -X PATCH localhost:8080/config -H 'Content-Type: application/json; charset=utf-8' -d '{"socks-port": "0", "port": "0" }'
```

oasiscifr:
@Dreamacro Great. So I can also call ApplyConfig and pass port=0 in the config to close them right?

Dreamacro:
@oasiscifr yep

oasiscifr:
@Dreamacro I've tried. The socks UDP listener didn't stopped.

oasiscifr:
I'll compose a PR to fix it.

PR #357, opened 2019-10-13
Optimize: fallback validTest

comwrg:


adapters/outbound/fallback.go
```diff
@@ -4,8 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"github.com/Dreamacro/clash/common/picker"
```
Dreamacro: clash follows a simple rule in `import`

```go
import (
    // go pkg

    // self pkg

    // third party pkg
)
```

PR #358, opened 2019-10-13
Feature(API): support mannually trigger urltest

tslling:
#273 
sorry, i made a mess, so i open this PR again here. 
new API: ```/proxies/:name/reelect```

request parameter
- name (required and case-sensitive): name of a urltest proxy group
- timeout (required): timeout for speedtest of urltest

response:
- now: current proxy after speed test


adapters/outbound/base.go
```diff
@@ -146,6 +146,12 @@ func (p *Proxy) MarshalJSON() ([]byte, error) {
 // URLTest get the delay for the specified URL
 func (p *Proxy) URLTest(ctx context.Context, url string) (t uint16, err error) {
 	defer func() {
+		if t == 0 {
```
Dreamacro: `ctx.Err() == context.Canceled` is better

hub/route/proxies.go
```diff
@@ -24,6 +24,7 @@ func proxyRouter() http.Handler {
 		r.Use(parseProxyName, findProxyByName)
 		r.Get("/", getProxy)
 		r.Get("/delay", getProxyDelay)
+		r.Get("/reelect", reelectProxy)
```
Dreamacro: I prefer `/healthcheck`

adapters/outbound/urltest.go
```diff
@@ -73,17 +73,22 @@ func (u *URLTest) Destroy() {
 	u.done <- struct{}{}
 }
 
+func (u *URLTest) SpeedTest(ctx context.Context) string {
```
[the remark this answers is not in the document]
tslling: Will each proxy implements ```HealthCheck```? if not, putting ```HealthCheck``` in ProxyGroup is better. But it seems that we did not have such a structure for ProxyGroup.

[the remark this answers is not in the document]
Dreamacro: `HealthCheck` could replace `SpeedTest`. For Proxy, It sends a request with a url and flushes the record, for Proxy Group, It also changes the selected node (e.g. url-test)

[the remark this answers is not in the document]
tslling: after trying for several times, I found ```Base``` is not a perfect place for ```HealthCheck```. I put it in ```ProxyAdapter``` now.  take a look

tslling:
If user requests to do a health check while this proxy is under check right now, we have two options.
* return an error indicating that this proxy is under health check now
* try periodically until dead line exceed

the first one is used now, is it acceptable?

Dreamacro:
@tslling `"golang.org/x/sync/singleflight"` is better

adapters/outbound/loadbalance.go
```diff
@@ -103,28 +104,46 @@ func (lb *LoadBalance) Destroy() {
...
-	wg := sync.WaitGroup{}
-	wg.Add(len(lb.proxies))
-
-	for _, p := range lb.proxies {
-		go func(p C.Proxy) {
-			p.URLTest(context.Background(), lb.rawURL)
-			wg.Done()
-		}(p)
+func (lb *LoadBalance) HealthCheck(ctx context.Context, url string) (uint16, error) {
+	if url == "" {
```
Dreamacro: is there url be empty?

adapters/outbound/fallback.go
```diff
@@ -98,24 +103,33 @@ func (f *Fallback) findAliveProxy() C.Proxy {
...
+		_, err := proxy.HealthCheck(ctx, url)
+		if err != nil {
+			return nil, err
+		}
+		return proxy, nil
+	}
+	select {
+	case <-ctx.Done():
+		return 0, errTimeout
+	case result := <-f.group.DoChan("healthcheck", func() (interface{}, error) {
```
Dreamacro: because of the name of proxy is unique, why not put a `singleflight.Group` to global, use DoChan or Do with name as the key. it can avoid the unnecessary request.

adapters/outbound/loadbalance.go
```diff
@@ -103,28 +104,46 @@ func (lb *LoadBalance) Destroy() {
...
+			return result.Val.(uint16), nil
+		}
+		return 0, result.Err
+	}
 }
 
 func (lb *LoadBalance) loop() {
 	tick := time.NewTicker(lb.interval)
-	go lb.validTest()
+	go func() {
```
Dreamacro: is timeout can passthrough to `healthCheck`?

hub/route/proxies.go
```diff
@@ -130,3 +131,41 @@ func getProxyDelay(w http.ResponseWriter, r *http.Request) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*time.Duration(timeout))
+	defer cancel()
+
+	t, err := proxy.HealthCheck(ctx, url)
+	if err != nil {
+		render.JSON(w, r, render.M{
+			"health": false,
+			"err":    err.Error(),
+		})
+	} else {
```
Dreamacro: maybe we don't need this else

adapters/outbound/loadbalance.go
tslling: I am considering that we already have a built-in ```url``` for proxy group in config file, so we can omit it in the request paramter: ```/proxy/:proxygroup/healthcheck?timeout=3000```, also the user-specified ```url``` will be respected.

adapters/outbound/fallback.go
tslling: using proxy name as key may not extend well, say, if we have other situations where singleflight is needed in the future. Basically we have two strategies:
* group of function(say healthcheck), then group name as key
* group of proxy group(say url-test, fallback), then function(healthcheck) as key

I am using the second one, you are suggesting the first one, right?


>it can avoid the unnecessary request.

I did not quite understand this.

Dreamacro: if I have a batch request for speedtest, A is contains in B, A do health check twice.

`${proxy.Name}:${url}:${timeout}` maybe a good key for `singleflight.Group`

adapters/outbound/loadbalance.go
tslling: ```ctx``` will be respected in ```groupHealthCheck```:
https://github.com/tslling/clash/blob/114c48c0716d24de0768ef0a77f441f201679321/adapters/outbound/util.go#L229-L233

hub/route/proxies.go
tslling: done

adapters/outbound/fallback.go
tslling: I see, it makes sense. But here is one thing worth mentioning.

A health check on a proxy group may cause checks on many single proxies, those checks on single proxy should not suppress user-requested check on same single proxy. Because those “affiliated” checks may get cancelled by fastest peer.

From this perspective, a global singleflight group will not save that much unnecessary requests.

Dreamacro: Is the previous behavior wrong? It's good to use `WithoutAutoCancel` on group healthcheck?

tslling: What I want to do is responding to user as quickly as possible(once the fastest proxy is found). But it seems that things are getting more complex than expected...
It's ok to do health check without canceling, in this way, we wont respond to user as quickly as possible(but at least within timeout)
So are you suggesting check without cancel?

Dreamacro: `WithoutAutoCancel` won't block the code: https://github.com/Dreamacro/clash/blob/e5284cf647717a8087a185d88d15a01096274bc2/common/picker/picker_test.go#L41

tslling: > ```WithoutAutoCancel``` won't block the code

but ```Wait``` will.
Can we just return once the fastest proxy is found and leave other proxies in testing process? I'm not sure if this is safe or not, because previously we wait here:
https://github.com/Dreamacro/clash/blob/82a8c03953a508bb59c1a33660de05d2f4c9cadb/adapters/outbound/urltest.go#L130-L135

Dreamacro: I think it's safe if it has a timeout context

tslling: For example, a health check request on proxy group A finished in 100ms, then it returns, but leaves other proxy(sa, proxy B) still in testing process.
Then user requested a health check on A again, and B happens to finished in 150ms, at this time, the second request get the wrong delay(50ms) for proxy B, and may consider B as the fastest proxy.

the key here is test in previous round will be reused in current round. To avoid this, we must not reuse previous test result. If so, neither singleflight nor the ```once``` protection is useful here(since it's safe to issue two conccurent test for a single proxy). We should probably move the singleflight at a much early stage(say, at the very beginning where we recieved the request).

Dreamacro: ok, I got it. There are two different cases:
* want to singleflight request
    * speedtest api
    * proxy group interval healthcheck
* don't want to singleflight request
    * select the fastest node

Do you have any idea about this?

tslling: It depends on what we want to do.
If we only focus on finding the fastest proxy for a group, we can cancel other testing. (as it is now)
Or, we can wait until all proxies finished their speed test or timeout, then iterate all proxy in this proxy group to get the fastest proxy using their latest delay. In this way, we allow multiple tests to  begin at different time, so that we can reuse speedtest result and handle the situation mentioned above correctly.

tslling: @Dreamacro hi, what should we do next?

Dreamacro: sorry, I'm a bit busy recently. I have to use a period of continuous-time for review. I am thinking of reducing the number of requests as much as possible. But now it seems that it can’t be implemented. I would review ASAP

Dreamacro:
Sorry for the long time no response. I'm working on a new feature called "Provider". It could read a lot of proxy from a file or an HTTP url. And `Proxy Group` could use it directly. For example:

```yaml
Proxy Group:
  - name: "fallback-auto"
    type: fallback
    use:
      - test
      - hk

proxy-provider:
  groupA:
    type: file
    path: ./groupA.yaml
    health-check:
      enable: true
      url: google.com
      interval: 300
  hk:
    type: http
    path: ./hk.yaml
    url: https://yourdoamin/group.yaml
    health-check:
      enable: true
      url: google.com
      interval: 300
```
so in proxy group adapter, the proxy group only use to pick proxy. And about field `proxies`, would wrapper with a provider. It's anything that I am not well considered? It seems like it would break this PR, so I want to hear your suggestion.

tslling:
@Dreamacro I am fine with that, this PR aims to provide an API to do healthcheck, as long as the new feature can get such an API done, it can definitely replace this PR.

And I am curious about the content of ```./groupA.yaml```, maybe discussed in a relevant PR/issue.  So should I close this PR now?

Howard-00:
> Sorry for the long time no response. I'm working on a new feature called "Provider". It could read a lot of proxy from a file or an HTTP url. And `Proxy Group` could use it directly. For example:
> 
> ```yaml
> Proxy Group:
>   - name: "fallback-auto"
>     type: fallback
>     use:
>       - test
>       - hk
> 
> proxy-provider:
>   groupA:
>     type: file
>     path: ./groupA.yaml
>     health-check:
>       enable: true
>       url: google.com
>       interval: 300
>   hk:
>     type: http
>     path: ./hk.yaml
>     url: https://yourdoamin/group.yaml
>     health-check:
>       enable: true
>       url: google.com
>       interval: 300
> ```
> 
> so in proxy group adapter, the proxy group only use to pick proxy. And about field `proxies`, would wrapper with a provider. It's anything that I am not well considered? It seems like it would break this PR, so I want to hear your sugges possible to update proxies automatically without restarting clash? 

> Sorry for the long time no response. I'm working on a new feature called "Provider". It could read a lot of proxy from a file or an HTTP url. And `Proxy Group` could use it directly. For example:
> 
> ```yaml
> Proxy Group:
>   - name: "fallback-auto"
>     type: fallback
>     use:
>       - test
>       - hk
> 
> proxy-provider:
>   groupA:
>     type: file
>     path: ./groupA.yaml
>     health-check:
>       enable: true
>       url: google.com
>       interval: 300
>   hk:
>     type: http
>     path: ./hk.yaml
>     url: https://yourdoamin/group.yaml
>     health-check:
>       enable: true
>       url: google.com
>       interval: 300
> ```
> 
> so in proxy group adapter, the proxy group only use to pick proxy. And about field `proxies`, would wrapper with a provider. It's anything that I am not well considered? It seems like it would break this PR, so I want to hear your suggestion.

Is possible to update proxies automatically without restarting clash?

Dreamacro:
@Howard-00 of course, it can 😆

Dreamacro:
Sorry for taking up your time, all side effect move to `provider`. Thanks for your contribution

PR #361, opened 2019-10-14
Some code improvements

sosiska:
* Rewrite if-else statements as switch.
Based on Go style guide:
https://golang.org/doc/effective_go.html#switch
* Some assignments simplified by using assignment operators.


Dreamacro:
thx for your contribution.

PR #429, opened 2019-12-11
Add support for freebsd/arm64

dmgk:
Go 1.14 will add support for freebsd/arm64, golang.org/x/net@83d349e8ac1a and golang.org/x/sys@33540a1f6037 are earliest commits that bring in necessary arm64 bits.


Dreamacro:
already update the dependencies at https://github.com/Dreamacro/clash/blob/dev/go.mod, which would merge to master when it ready.

thanks for your pr, but I can't merge it.

yurivict:
@Dreamacro  FYI with the current dependencies the FreeBSD port breaks on arm64.

golang:sys is too old.

This PR needs to be merged.

PR #480, opened 2020-01-12
fix cache invalid proxy

comwrg:
If during cache time, proxy not alive, the `u.fastSingle.Do` will still return invalid proxy.

### DIFF
--- a/adapters/outboundgroup/urltest.go
+++ b/adapters/outboundgroup/urltest.go
@@ -48,7 +48,7 @@ func (u *URLTest) proxies() []C.Proxy {
 }
 
 func (u *URLTest) fast() C.Proxy {
-	elm, _, _ := u.fastSingle.Do(func() (interface{}, error) {
+	fn := func() (interface{}, error) {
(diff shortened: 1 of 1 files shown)


Dreamacro:
`fastSingle` cache item for 10s, I don't think it's necessary

comwrg:
If not check alive flag, cache time decrease to 5s is better, due to default timeout is 5s.

adapters/outboundgroup/urltest.go
```diff
@@ -64,7 +64,12 @@ func (u *URLTest) fast() C.Proxy {
 			}
 		}
 		return fast, nil
-	})
+	}
+
+	elm, _, _ := u.fastSingle.Do(fn)
+	if !elm.(C.Proxy).Alive() {
+		elm, _ = fn()
```
Dreamacro: sorry for the delay, it causes another problem: when `fastest` proxy dead, it always goes this path until the cache expired.

I think 10s is enough for url-test

PR #489, opened 2020-01-27
Chore: improve Dockerfile, make the docker image can be use in Raspberry Pi

bolasblack:
So we can use the docker image in Raspberry Pi or other arch like

    docker run --name clash dreamacro/clash /clash-linux-armv7

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,12 +5,12 @@ RUN apk add --no-cache make git && \
 WORKDIR /clash-src
 COPY . /clash-src
 RUN go mod download && \
-    make linux-amd64 && \
-    mv ./bin/clash-linux-amd64 /clash
+    make all-arch
 
 FROM alpine:latest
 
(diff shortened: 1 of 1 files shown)


Dreamacro:
It makes images large. I push `Dockerfile.arm32v7` and `Dockerfile.arm64v8`, but It doesn't build automated now.

bolasblack:
@Dreamacro Great!

Maybe you can trigger build manually like this?

图片

图片

Dreamacro:
@bolasblack Yes, I have tried to use docker hub automated build, but it built failed.

image

It seems that missing `qemu-user-static` (https://stackoverflow.com/questions/54578066/how-to-build-a-docker-image-on-a-specific-architecture-with-docker-hub). I would do it later.

bolasblack:
Thank you so much!

Dreamacro:
close because have another solution

PR #524, opened 2020-02-21
Feature: add configuration test

whtsky:
I generate clash config with [clash-config-preprocessor](https://github.com/whtsky/clash-config-preprocessor/). Sometimes the generated config file is invalid due to upstream rule-set changes, so it would be nice to have the configuration file test feature built-in.

The cli design is highly influenced by `nginx -t`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -63,12 +63,11 @@ func main() {
 
 	if testConfig {
 		if _, err := executor.Parse(); err != nil {
-			fmt.Fprintln(os.Stderr, err.Error())
+			log.Errorln(err.Error())
 			fmt.Printf("configuration file %s test failed\n", constant.Path.Config())
 			os.Exit(1)
-		} else {
-			fmt.Printf("configuration file %s test is successful\n", constant.Path.Config())
 		}
+		fmt.Printf("configuration file %s test is successful\n", constant.Path.Config())
 		return
 	}
 
--- a/main.go
+++ b/main.go
@@ -62,8 +62,6 @@ func main() {
 	}
 
 	if testConfig {
-		// silence INFO logs during config parsing
-		log.SetLevel(log.WARNING)
 		if _, err := executor.Parse(); err != nil {
 			fmt.Fprintln(os.Stderr, err.Error())
 			fmt.Printf("configuration file %s test failed\n", constant.Path.Config())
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -57,6 +61,19 @@ func main() {
 		log.Fatalln("Initial configuration directory error: %s", err.Error())
 	}
 
+	if testConfig {
+		// silence INFO logs during config parsing
+		log.SetLevel(log.WARNING)
```
Dreamacro: why not display provider log

whtsky: IMO display `Start initial compatible provider XXX` during configuration test is confusing.

https://github.com/Dreamacro/clash/blob/609869bf5a1d779b0ac6a0caca9ff9ea67b0c40f/config/config.go#L331

Dreamacro: but it does a part of the parse.

whtsky: 图片

whtsky: removed

main.go
```diff
@@ -57,6 +61,17 @@ func main() {
 		log.Fatalln("Initial configuration directory error: %s", err.Error())
 	}
 
+	if testConfig {
+		if _, err := executor.Parse(); err != nil {
+			fmt.Fprintln(os.Stderr, err.Error())
```
Dreamacro: I prefer all logs use `"github.com/Dreamacro/clash/log"`

if you want to split `warning` `error` into stderr, you can refer https://github.com/sirupsen/logrus/issues/403 and modify `"github.com/Dreamacro/clash/log"`

whtsky: Ok, I changed the error output to use log.

As for `configuration file %s test failed | success`, I believe they are not part of log but rather standard outputs - like `clash -v` and what nginx does:
图片

Dreamacro: fine

main.go
```diff
@@ -57,6 +61,17 @@ func main() {
 		log.Fatalln("Initial configuration directory error: %s", err.Error())
 	}
 
+	if testConfig {
+		if _, err := executor.Parse(); err != nil {
+			log.Errorln(err.Error())
+			fmt.Printf("configuration file %s test failed\n", constant.Path.Config())
+			os.Exit(1)
+		} else {
```
Dreamacro: The last problem: the `else` is not necessary

Dreamacro:
LGTM, Thanks 👍

PR #542, opened 2020-02-28
feature: allow multiple ip for one host

ManassehZhou:
#382 
modified the type of hosts to `map[string][]string`
used a time-based round-robin to determine which result should be resolved

### DIFF
--- a/component/domain-trie/trie_test.go
+++ b/component/domain-trie/trie_test.go
@@ -3,9 +3,11 @@ package trie
 import (
 	"net"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
-var localIP = net.IP{127, 0, 0, 1}
+var localIP = []net.IP{{127, 0, 0, 1}, {127, 0, 0, 2}}
 
 func TestTrie_Basic(t *testing.T) {
 	tree := New()
@@ -23,9 +25,7 @@ func TestTrie_Basic(t *testing.T) {
 		t.Error("should not recv nil")
 	}
 
-	if !node.Data.(net.IP).Equal(localIP) {
-		t.Error("should equal 127.0.0.1")
-	}
+	assert.Equal(t, localIP, node.Data.([]net.IP), "should be the same IP addresses")
 
 	if tree.Insert("", localIP) == nil {
 		t.Error("should return error")
--- a/component/resolver/resolver.go
+++ b/component/resolver/resolver.go
@@ -6,7 +6,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Dreamacro/clash/common/cache"
 	trie "github.com/Dreamacro/clash/component/domain-trie"
+	"github.com/Dreamacro/clash/constant"
 )
 
 var (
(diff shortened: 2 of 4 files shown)


Dreamacro:
Can you explain why you need multiple IPs in one host?

ManassehZhou:
When developing a web server, you may deploy the same software to different servers but with one domain, with a host can be resolved as multiple IPs, you can test whether the distributed system works or not.

Dreamacro:
In my experience, web servers are behind a load-balance like nginx or caddy.

> with a host can be resolved as multiple IPs, you can test whether the distributed system works or not.

It seems that IP change in the 30s. is it inconvenient?

soffchen:
I think Hosts is equal to /etc/hosts, so it should follow its original logic: single ip -> single hostname.

ManassehZhou:
> I think Hosts is equal to /etc/hosts, so it should follow its original logic: single ip -> single hostname.

By editing /etc/hosts, it will only return the first record instead fo change into different ones.





> In my experience, web servers are behind a load-balance like nginx or caddy.

sure it is, but it could also be an environment without EIP, i think one host to multi IPs method is more simple.

> It seems that IP change in the 30s. is it inconvenient?

maybe leave it for customization?

BirkhoffLee:
There are loads of websites that load balances first at the DNS level, just grabbed one:

```
$ dig www.ikea.com.tw
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 21451
;; Flags: qr rd ra; QUERY: 1; ANSWER: 5; AUTHORITY: 0; ADDITIONAL: 0

;; QUESTION SECTION:
;; www.ikea.com.tw.             IN      A

;; ANSWER SECTION:
www.ikea.com.tw.        3600    IN      CNAME   d1j3z7hirljnop.cloudfront.net.
d1j3z7hirljnop.cloudfront.net.  60      IN      A       13.35.67.14
d1j3z7hirljnop.cloudfront.net.  60      IN      A       13.35.67.118
d1j3z7hirljnop.cloudfront.net.  60      IN      A       13.35.67.123
d1j3z7hirljnop.cloudfront.net.  60      IN      A       13.35.67.13

;; Received 140 B
;; Time 2020-02-29 01:01:06 CST
;; From 1.1.1.1@53(UDP) in 145.7 ms
```

Dreamacro:
```yaml
hosts:
  '*.clash.dev':
    - 127.0.0.1
  'alpha.clash.dev':
    - '::1'
```

it's not friendly for users

ManassehZhou:
> ```yaml
> hosts:
>   '*.clash.dev':
>     - 127.0.0.1
>   'alpha.clash.dev':
>     - '::1'
> ```
> 
> it's not friendly for users

so just put all the ip address into a string splitted by comma?

xfoxfu:
Maybe we can detect whether the value is an array or a string, and parse it always into an array by procedures in <https://github.com/go-yaml/yaml/issues/100>.

fishg:
这个特性挺好啊

PR #562, opened 2020-03-07
[Feature] Add UDP TPROXY support for redir on Linux

duament:
#234 #387

#### Usage
See `ss-redir`: https://github.com/shadowsocks/shadowsocks-libev#transparent-proxy

PR #572, opened 2020-03-09
Fix UDP NAT type for TPROXY

duament:
This pr is inspired by 9010090800663aa0449e88888b9ede8ffb3361b6. It improves UDP NAT behavior for TPROXY(#562).
### details
In `WriteBack` function, the socket should bind to `origDst` only if `origDst` maps to a host(fake-ip or redir-host mode). Otherwise it'll be ok to bind to `addr`(the real source of the UDP packet to write back).

### DIFF
--- a/proxy/redir/udp.go
+++ b/proxy/redir/udp.go
@@ -35,8 +35,7 @@ func NewRedirUDPProxy(addr string) (*RedirUDPListener, error) {
 		oob := make([]byte, 1024)
 		for {
 			buf := pool.BufPool.Get().([]byte)
-
-			n, oobn, _, remoteAddr, err := c.ReadMsgUDP(buf, oob)
+			n, oobn, _, lAddr, err := c.ReadMsgUDP(buf, oob)
 			if err != nil {
 				pool.BufPool.Put(buf[:cap(buf)])
 				if rl.closed {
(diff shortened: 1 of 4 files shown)


Dreamacro:
I don't think this can improve NAT type. The Fullcone NAT type could receive different net.Addr packet. And the caller of `WriteBack` already pass the right addr to fakeConn

duament:
> And the caller of `WriteBack` already pass the right addr to fakeConn

Actually, #562 always use `origDst` instead of `addr` due to a fake-ip issue.

Dreamacro:
`resolver.DefaultResolver.(*dns.Resolver)` is not safe. resolver can export a function IsFakeIP so that can pass fakeip into fakeConn

comzyh:
> `resolver.DefaultResolver.(*dns.Resolver)` is not safe. resolver can export a function IsFakeIP so that can pass fakeip into fakeConn

Looking forward to this change.

However, I still suggest that `channel` take the responsibility to determine if the IP is mapped. Then UDP inbounds (redir, socks5, Tun) don't need to write its own Fake-IP judger.

Ref:

https://github.com/Dreamacro/clash/blob/609869bf5a1d779b0ac6a0caca9ff9ea67b0c40f/constant/adapters.go#L119-L123

Dreamacro:
@duament @comzyh cc af31e35

duament:
I think `IPToHost` is better than `IsFakeIP`. Consider such a case where `dns.google` maps to `8.8.8.8` in clash but the remote server connects to `8.8.4.4`.
```
➜  ~ dig @dns.google www.google.com
;; reply from unexpected source: 8.8.4.4#53, expected 8.8.8.8#53
```

Dreamacro:
This is the limitation of udp, some protocol like ss support remote resolve ip. it returns ip instead of domain

duament:
Fine

comzyh:
@Dreamacro af31e35 looks good to me. Thanks.

PR #581, opened 2020-03-15
Fix: config MarshalYAML

kamingchan:


log/level.go
```diff
@@ -55,6 +55,11 @@ func (l LogLevel) MarshalJSON() ([]byte, error) {
 	return json.Marshal(l.String())
 }
 
+// MarshalYAML serialize LogLevel with yaml
+func (l LogLevel) MarshalYAML() (interface{}, error) {
```
Dreamacro: It's not necessary to  serialize `LogLevel` and `TunnelMode` into yaml

dns/util.go
```diff
@@ -46,8 +45,8 @@ func (e *EnhancedMode) UnmarshalYAML(unmarshal func(interface{}) error) error {
 }
 
 // MarshalYAML serialize EnhancedMode with yaml
-func (e EnhancedMode) MarshalYAML() ([]byte, error) {
-	return yaml.Marshal(e.String())
+func (e EnhancedMode) MarshalYAML() (interface{}, error) {
```
Dreamacro: https://godoc.org/gopkg.in/yaml.v2#Marshal

kamingchan: https://pkg.go.dev/gopkg.in/yaml.v2?tab=doc#Marshaler

log/level.go
kamingchan: My program import `github.com/Dreamacro/clash` and use `config.RawDNS` to build my own clash config. Code below shows how I use it.

```go
package main

import (
	"github.com/Dreamacro/clash/config"
	"github.com/Dreamacro/clash/dns"
	"github.com/Dreamacro/clash/log"
	"github.com/Dreamacro/clash/tunnel"
	"gopkg.in/yaml.v2"
	"os"
)

func main() {
	yaml.NewEncoder(os.Stdout).Encode(&config.RawConfig{
		Mode:     tunnel.Rule,
		LogLevel: log.INFO,
		DNS: config.RawDNS{
			EnhancedMode: dns.FAKEIP,
		},
	})
}
```

However, none of them implements `MarshalYAML() (interface{}, error)` and the output will be

```yaml
mode: 1
log-level: 1
dns:
  enhanced-mode: 1
```

which is not recognized by clash core.

 In this PR, marshal a `config.RawConfig` would be like

```yaml
mode: Rule
log-level: info
dns:
  enhanced-mode: fake-ip
```

Dreamacro: fine

PR #597, opened 2020-03-23
Feature: support proxy-group in relay

duament:
### example

```yaml
proxy-groups:
- name: "select"
  type: select
  proxies:
    - ss1
    - ss2

- name: "relay"
  type: relay
  proxies:
    - select
    - vmess1
```

### known issues

- Healthchecks don't follow the proxy chains.

### DIFF
--- a/adapters/outbound/base.go
+++ b/adapters/outbound/base.go
@@ -53,7 +53,7 @@ func (b *Base) Addr() string {
 	return b.addr
 }
 
-func (b *Base) Proxy(metadata *C.Metadata) C.Proxy {
+func (b *Base) Unwrap(metadata *C.Metadata) C.Proxy {
 	return nil
 }
 
--- a/adapters/outbound/base.go
+++ b/adapters/outbound/base.go
@@ -53,6 +53,10 @@ func (b *Base) Addr() string {
 	return b.addr
 }
 
+func (b *Base) Proxy() C.Proxy {
+	return nil
+}
+
 func NewBase(name string, addr string, tp C.AdapterType, udp bool) *Base {
 	return &Base{name, addr, tp, udp}
 }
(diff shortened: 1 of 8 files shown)


Dreamacro:
why not Implementation `StreamConn` on group?

duament:
> why not Implementation `StreamConn` on group?

Because the proxy may be changed between calling of `Addr` and `StreamConn`

Dreamacro:
LoadBalance SHOULD NOT be a round-robin or random strategy, it's a bad practice on proxy application

duament:
> LoadBalance SHOULD NOT be a round-robin or random strategy, it's a bad practice on proxy application

This is not true if the endpoint of proxy chains is a specific one

Dreamacro:
> This is not true if the endpoint of proxy chains is a specific one

What if it is last node?

duament:
I fixed load-balance by adding a new parameter `metadata` to `Proxy()`

h0cheung:
Thanks a lot. What makes it difficult to relay UDP?

adapters/outbound/base.go
```diff
@@ -53,6 +53,10 @@ func (b *Base) Addr() string {
 	return b.addr
 }
 
+func (b *Base) Proxy(metadata *C.Metadata) C.Proxy {
```
Dreamacro, line 56: The method name is ambiguous. maybe called it `UnWrap` or some other name. If the `plain proxy` return itself,  is it better?

constant/adapters.go
```diff
@@ -69,6 +69,7 @@ type ProxyAdapter interface {
 	SupportUDP() bool
 	MarshalJSON() ([]byte, error)
 	Addr() string
+	Proxy(metadata *Metadata) Proxy
```
Dreamacro, line 72: Is it put this to interface `Proxy`? It would be better if there were some comments here

adapters/outbound/base.go
duament: I think it's better to return nil, just like what [errors.Unwrap](https://golang.org/pkg/errors/#Unwrap) does.

PR #630, opened 2020-04-10
[Fix] Set SO_REUSEADDR for UDP listeners on linux

duament:
Fix issue #616

### DIFF
--- a/common/sockopt/reuseaddr_linux.go
+++ b/common/sockopt/reuseaddr_linux.go
@@ -1,5 +1,3 @@
-// +build linux
-
 package sockopt
 
 import (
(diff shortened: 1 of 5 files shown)


common/sockopt/reuseaddr_linux.go
```diff
@@ -0,0 +1,21 @@
+// +build linux
```
Dreamacro, line 1: *_linux.go don't need this comment

common/sockopt/reuseaddr_linux.go
```diff
@@ -0,0 +1,21 @@
...
+
+package sockopt
+
+import (
+	"net"
+	"syscall"
+)
+
+func Reuseaddr(c *net.UDPConn) (err error) {
```
Dreamacro, line 10: maybe ReuseUDPConnAddr ?

dns/server.go
```diff
@@ -4,6 +4,8 @@ import (
 	"net"
 
 	D "github.com/miekg/dns"
+
+	"github.com/Dreamacro/clash/common/sockopt"
```
Dreamacro, line 8: std lib

internal path

third party lib

PR #638, opened 2020-04-17
Fix: dns tcp-tls X509.HostnameError

tete1030:
DoT with hostname instead of bare IP has never been working. This problem normally would be covered up when multiple DNS servers presented. If only one DoT DNS server with hostname, e.g. "tls://dns.rubyfish.cn:853", the log would show "All DNS requests failed". 

My config:
```yaml
dns:
  enhanced-mode: redir-host
  listen: 127.0.0.1:7874
  enable: true
  ipv6: false
  default-nameserver:
  - 114.114.114.114
  nameserver:
  - 114.114.114.114
  fallback:
  - tls://dns.rubyfish.cn:853
  fallback-filter:
    geoip: true
    ipcidr:
      - 240.0.0.0/4
      - 0.0.0.0/32
```

Adding the following debug logger to `dns/client.go`:
image

the root problem can be found when `dig @192.168.1.1 www.google.com` (my clash running in x86-64 router):
image

The problem is obvious, 'miekg/dns' has never been informed the hostname but only a resolved IP. After this patch applied:
image

Edit:
I just found the `dev` branch has already been showing first DNS error, not in `master`. So the logger I added is not necessary to observe this problem

### DIFF
--- a/dns/util.go
+++ b/dns/util.go
@@ -133,6 +133,7 @@ func transform(servers []NameServer, resolver *Resolver) []dnsClient {
 					ClientSessionCache: globalSessionCache,
 					// alpn identifier, see https://tools.ietf.org/html/draft-hoffman-dprive-dns-tls-alpn-00#page-6
 					NextProtos: []string{"dns"},
+					ServerName: host,
 				},
 				UDPSize: 4096,
 				Timeout: 5 * time.Second,


Dreamacro:
Thanks

tete1030:
😄被这个问题迷惑了昨天一下午，到半夜才找到根源

PR #645, opened 2020-04-20
Improve: add basic auth support for provider URL

ruisiji:
Copy from https://github.com/Kr328/ClashForAndroid/commit/c5104c241a1a38f3d46339250b6f3d7bd96937e9


Dreamacro:
thanks for patch

PR #647, opened 2020-04-22
[Feature] DNS: employ LruCache for optimistic DNS

comzyh:
This PR replaces the existing DNS cache with LRU Cache and thus can achieve optimistic DNS.

### Changes:

When a DNS query comes to clash, if there is an expired msg in cache match the query, clash will return the expired msg and start to fetch the newer response from upstream **async**.

When mapping(redir-host/fake-ip) is enabled, when request without Host but IP comes to clash, clash will do IP-to-Host checking using this IP in LRU Cache. This procedure **will not** affect the TTL of the corresponding DNS messages.

### Discuss:

- What TTL should clash reply when an expired DNS msg return?
- How large should the cache capacity be set?
- Should cache size and optimistic DNS be configurable?

### DIFF
--- a/common/cache/lrucache.go
+++ b/common/cache/lrucache.go
@@ -42,6 +42,13 @@ func WithSize(maxSize int) Option {
 	}
 }
 
+// WithStale decide whether Stale return is enabled.
+func WithStale(stale bool) Option {
+	return func(l *LruCache) {
+		l.staleReturn = stale
+	}
+}
+
 // LruCache is a thread-safe, in-memory lru-cache that evicts the
 // least recently used entries from memory when (if set) the entries are
 // older than maxAge (in seconds).  Use the New constructor to create one.
@@ -52,6 +59,7 @@ type LruCache struct {
 	cache          map[interface{}]*list.Element
 	lru            *list.List // Front is least-recent
 	updateAgeOnGet bool
+	staleReturn    bool
 	onEvict        EvictCallback
 }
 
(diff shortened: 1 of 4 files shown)


fedstryale:
I compiled with this code and everything now seems to work well but I have  two questions:

1. Will optimistic dns now work with clash to retry connection when msg is expired AND connection cannot be established for a certain time?
2. Will optimistic dns still work when socks5 requests(Hosts or IPs) are received?

comzyh:
1. No. In this PR, clash will not try to reconnect remote host if the old host is really out of service. But I think most of the client has it's own reconnect procedure. In this PR, clash will return stale DNS record with TTL=1, if a client wants to reconnect, I guess it will get an up-to-date response.

2. it depends. For socks5 requests with a hostname in its header, clash will resolve it's IP address only when the request encounters a **IP-based rule**, or the final outbound doesn't support Hostname(like DIRECT), otherwise, clash will pass this hostname to remote proxy server.
  If clash does try to resolve this hostname, the optimistic will take effect.

Dreamacro:
add test plz

common/cache/lrucache.go
```diff
@@ -72,29 +81,26 @@ func NewLRUCache(options ...Option) *LruCache {
...
-	}
+	return value, true
+}
 
-	c.lru.MoveToBack(le)
-	entry := le.Value.(*entry)
-	if c.maxAge > 0 && c.updateAgeOnGet {
-		entry.expires = time.Now().Unix() + c.maxAge
+// GetWithExpire returns the interface{} representation of a cached response,
+// a int64 give the expected expires timestamp in seconds
```
Dreamacro, line 94: You may have forgotten to change this

common/cache/lrucache.go
```diff
@@ -72,29 +81,26 @@ func NewLRUCache(options ...Option) *LruCache {
...
+// and  a bool set to true if the key was found.
```
Dreamacro, line 95: An extra space lol

dns/resolver.go
```diff
@@ -96,22 +96,36 @@ func (r *Resolver) Exchange(m *D.Msg) (msg *D.Msg, err error) {
...
+		if expireTime.Before(now) {
+			setMsgTTL(msg, uint32(1)) // Continue fetch
+			go r.exchangeWithoutCache(m)
+		} else {
+			setMsgTTL(msg, uint32(expireTime.Sub(time.Now()).Seconds()))
+		}
 		return
 	}
+	return r.exchangeWithoutCache(m)
+
```
Dreamacro, line 112: empty line aha

comzyh:
Thanks, fixed.

Issue #531, opened 2020-02-23
[Feature Request] Add command-line options to override `external-controller`, `secret` and `external-ui`


ysc3839:
感谢你向 Clash Core 提交 Feature Request！
在提交之前，请确认：

- [x] 我已经在 [Issue Tracker](……/) 中找过我要提出的请求

请注意，如果你并没有遵照这个 issue template 填写内容，我们将直接关闭这个 issue。

我都确认过了，我要继续提交。

------------------------------------------------------------------

请附上任何可以帮助我们解决这个问题的信息，如果我们收到的信息不足，我们将对这个 issue 加上 *Needs more information* 标记并在收到更多资讯之前关闭 issue。

### Clash core config

```
……
```

### Clash log

```
……
```

### 环境 Environment

* Clash Core 的操作系统 (the OS that the Clash core is running on)
……
* 使用者的操作系统 (the OS running on the client)
……
* 网路环境或拓扑 (network conditions/topology)
……
* iptables，如果适用 (if applicable)
……
* ISP 有没有进行 DNS 污染 (is your ISP performing DNS pollution?)
……
* 其他
……

### 说明 Description

Clash GUI like [ClashX](https://github.com/yichengchen/clashX) requires setting its own `external-ui` path and obtaining the port of `external-controller` so that it can communicate with Clash.
However currently these options can only be defined in `config.yaml`. And ClashX's workaround is [modify](https://github.com/yichengchen/clashX/blob/09ef15b285024f008d1c4ab23bbed69e5d7f05f4/ClashX/main.go) Clash, expose these options as library functions, and run Clash in-process.
I'm writing a Clash GUI, but I want to run Clash in a separate process. These command-line options are necessary.

### 更多信息 More Information
Related issue: #344

BirkhoffLee:
This will be a quality-of-life improvement for Clash. I'm as of now using managed configuration files from my proxy provider, and when I want to deploy Clash on my routers, I need to patch 'em up first, for example the `ports` or `external-controller`. Having CLI options to override the values in the configuration file is absolutely a must, and is what other software would've done.

ysc3839:
Another solution is spilt basic config and proxy config into two files (#306). This solves these two issues.

Dreamacro:
merged

PR #656, opened 2020-04-27
Feature: add command-line options to override `external-controller`, `secret` and `external-ui` (#531)

ysc3839:
Discussion: #531

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,18 +18,30 @@ import (
 )
 
 var (
-	version    bool
-	testConfig bool
-	homeDir    string
-	configFile string
+	flagset            map[string]bool
+	version            bool
+	testConfig         bool
+	homeDir            string
+	configFile         string
+	externalUI         string
+	externalController string
+	secret             string
 )
 
 func init() {
 	flag.StringVar(&homeDir, "d", "", "set configuration directory")
 	flag.StringVar(&configFile, "f", "", "specify configuration file")
+	flag.StringVar(&externalUI, "ext-ui", "", "override external ui directory")
+	flag.StringVar(&externalController, "ext-ctl", "", "override external controller address")
+	flag.StringVar(&secret, "secret", "", "override secret for RESTful API")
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -18,18 +18,30 @@ import (
...
 	flag.StringVar(&homeDir, "d", "", "set configuration directory")
 	flag.StringVar(&configFile, "f", "", "specify configuration file")
+	flag.StringVar(&externalUI, "ext-ui", "", "override external ui directory")
+	flag.StringVar(&externalController, "ext-ctl", "", "override external controller address")
+	flag.StringVar(&secret, "secret", "", "override secret for RESTful API")
 	flag.BoolVar(&version, "v", false, "show current version of clash")
 	flag.BoolVar(&testConfig, "t", false, "test configuration and exit")
 	flag.Parse()
+
+	flagset = make(map[string]bool)
```
Dreamacro, line 41: maybe we don't need this

ysc3839: So without this how to check if a flag exist?

Dreamacro: In this case, a not empty string is enough

ysc3839: But empty string can still be used to override config. For example: force an empty secret.

Dreamacro: what about `flag.Lookup`?

ysc3839: Not working, `flag.Lookup("ext-ui")` always returns not nil value.
```
fmt.Printf("ext-ui %t %t\n", flag.Lookup("ext-ui") != nil, flagset["ext-ui"])
```
```
$ ./clash
ext-ui true false
```
```
$ ./clash -ext-ui test
ext-ui true true
```

Dreamacro: ok, I didn't expect this library so sucks. recommend to use `map[string]bool{}` and format the code with `go fmt`

ysc3839: Updated.

Dreamacro:
👍Thanks for your contribution

PR #673, opened 2020-05-05
Optimize: Cache dns before timing duration urltest

comwrg:
Cache dns record before timing in order to ignore the impact of DNS time on the final results. #426

### DIFF
--- a/adapters/outbound/base.go
+++ b/adapters/outbound/base.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/Dreamacro/clash/common/queue"
+	"github.com/Dreamacro/clash/component/resolver"
 	C "github.com/Dreamacro/clash/constant"
 )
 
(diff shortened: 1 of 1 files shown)


Dreamacro:
These changes only effect the DIRECT but without other proxy nodes. This is unfair to other proxies. And other proxies don't need to resolve host before dial so that they can find the best ip

comwrg:
These changes will effect `URLTest` result. 
But you said
> These changes only effect the DIRECT but without other proxy nodes.

I have a little confused.

Dreamacro:
For DIRECT, dial a host could use the DNS cache directly, but in proxy node, clash would send the host (if the request has a host) and port.

comwrg:
https://github.com/Dreamacro/clash/blob/6043ad06bf7b04a93863303f38ab150d1cb283fd/adapters/outbound/base.go#L179

DNS query host is proxy's host not url's host.
The format of proxy address could be `domain:port` or `ip:port`

`domain:port` delay = query dns time + dial time
`ip:port`  delay = dial time

Dreamacro:
but why not put `start := time.Now()` under `instance, err := p.DialContext(ctx, &addr)` 🤔

Dreamacro:
The implementation is too hack, and only effect ipv4

comwrg:
yes, but I have no idea about how to calculate DNS query time. It is a little difficult to separate DNS and Dial code.

https://github.com/Dreamacro/clash/blob/c0a2473160f9da2ea2116358fed33a3e6a77ae5a/component/dialer/dialer.go#L78-L154

Dreamacro:
@comwrg extra dns time has improved after #647 merged

PR #839, opened 2020-07-29
Improve: simplify darwin process searching, read pid as native endian uint32

icpz:
+ Fix: xsocket_n.so_last_pid has native byte order

+ Improve: simplify pcblist parsing, reduce magic numbers


Dreamacro:
LGTM, thanks!

PR #840, opened 2020-07-29
Feature: PROCESS-NAME for windows

icpz:
只在虚拟机里 (windows 10 1909 amd64) 测试通过- -

大佬们帮忙测试一下其它版本的系统？

### DIFF
--- a/rules/process_windows.go
+++ b/rules/process_windows.go
@@ -34,38 +34,46 @@ var (
 	queryProcName uintptr
 
 	errNotFound = errors.New("process not found")
+
+	matchMeta = func(p *Process, m *C.Metadata) bool { return false }
 )
 
 func init() {
-	h, err := windows.LoadLibrary("iphlpapi.dll")
-	if err != nil {
-		log.Errorln("LoadLibrary iphlpapi.dll error: %s", err.Error())
-		return
-	}
+	err := func() error {
+		h, err := windows.LoadLibrary("iphlpapi.dll")
+		if err != nil {
+			return fmt.Errorf("LoadLibrary iphlpapi.dll failed: %s", err.Error())
+		}
 
-	getExTcpTable, err = windows.GetProcAddress(h, tcpTableFunc)
-	if err != nil {
-		log.Errorln("GetProcAddress of %s error: %s", tcpTableFunc, err.Error())
-		return
-	}
+		getExTcpTable, err = windows.GetProcAddress(h, tcpTableFunc)
+		if err != nil {
+			return fmt.Errorf("GetProcAddress of %s failed: %s", tcpTableFunc, err.Error())
+		}
 
-	getExUdpTable, err = windows.GetProcAddress(h, udpTableFunc)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


rules/process_windows.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+func getExecPathFromPID(pid uint32) (string, error) {
+	switch pid {
+	case 0:
+		// reserved pid for system idle process
+		return "", nil
+	case 4:
+		// reserved pid for windows kernel image
+		return "ntoskrnl.exe", nil
```
ysc3839, line 264: I think here should be `System`, and the above one should be `System Idle Process`.
Because it's standard name in Task Manager. And an empty name or a nonexist name (there's no `ntoskrnl.exe` in process list) may confuse users.

icpz: "System" could be a regular process name while WSL enabled.

Actually pid 0 doesn't need to be handled here, because the established TCP connections won't be owned by the idle process.

ysc3839: > "System" could be a regular process name while WSL enabled.

"System" could be a regular process name while WSL disabled.
image
`QueryFullProcessImageNameW` will return the full filesystem path for the process, not just a "process name". So it's enough to distinguish.

> Actually pid 0 doesn't need to be handled here, because the established TCP connections won't be owned by the idle process.

You are right.

icpz: > "System" could be a regular process name while WSL disabled.
> image
> `QueryFullProcessImageNameW` will return the full filesystem path for the process, not just a "process name". So it's enough to distinguish.
> 

What if the user only wants to match  the regular "system" process, but not the kernel itself?

I known `QueryFullProcessImageNameW` will return the full fs path, but  a PROCESS-NAME rule only check its basename. Under such a situation, returning "ntoskrnl.exe" for PID 4 is more precise and consistent with the behaviors of other normal PIDs since it's the actual basename of the path of windows kernel image.

ysc3839: So what if the user only wants to match the regular "ntoskrnl.exe" process, but not the kernel itself?

icpz: > So what if the user only wants to match the regular "ntoskrnl.exe" process, but not the kernel itself?

It's impossible. Say the user also can't only match a process with path "C:\a\p.exe" but not a "C:\b\p.exe", since both basenames are "p.exe". "ntoskrnl.exe" has made no difference with "p.exe" here.

icpz: > Another solution is to put a ":" in the kernel process name, for example `:System`. So no process will conflict with it.

Sounds great!

Dreamacro:
LGTM, thanks

debao1996lu:
tap不支持啊

icpz:
> tap不支持啊

建议使用 tun 模式。tap 模式并不是 core 这边的功能

debao1996lu:
> > tap不支持啊
> 
> 建议使用 tun 模式。tap 模式并不是 core 这边的功能

好的，没注意看，我去别的地方发

PR #848, opened 2020-07-30
Fix: handle snell server reported error message properly

icpz:
There is one byte error code before the length of message.

https://github.com/icpz/snell-server-reversed/blob/master/README.v1.md#s-to-c


Dreamacro:
LGTM, thanks

PR #852, opened 2020-07-30
Fix: dns server should lookup hosts

Kr328:
Fix https://github.com/Dreamacro/clash/issues/849  https://github.com/Dreamacro/clash/issues/759  https://github.com/Kr328/ClashForAndroid/issues/553 

不过这样处理一次访问似乎会 **查找两次 hosts**


Dreamacro:
prefer to add an option and use `withHosts`

PR #874, opened 2020-08-07
Improve: add `full-match' parameter to PROCESS-NAME

icpz:
This PR enables PROCESS-NAME to match the absolute path of a process executable, which will be of help to distinguish between different paths with same base names.

Example:
```yaml
rules:
- PROCESS-NAME,/usr/local/bin/url,DIRECT,full-match
- PROCESS-NAME,/usr/bin/curl,REJECT,full-match
- PROCESS-NAME,C:\Users\admin\Desktop\curl\bin\curl.exe,DIRECT,full-match
- PROCESS-NAME,curl,DIRECT
- PROCESS-NAME,/usr/bin/curl,DIRECT # same as above line, the process is reset to its base name if full-match omits
```

For consistency among all supported platforms, I changed the behavior on linux from `read(/proc/${PID}/cmdline)[0]` to `readlink(/proc/${PID}/exe)`. Please let me know if there are any side effects. @Kr328 Thanks!

MISC: The common parts are moved to `rules/process.go`.

### DIFF
--- a/rules/base.go
+++ b/rules/base.go
@@ -11,11 +11,21 @@ var (
 	ErrInvalidNetwork     = errors.New("invalid network")
 
 	noResolve = "no-resolve"
+	fullMatch = "full-match"
 )
 
 func HasNoResolve(params []string) bool {
 	for _, p := range params {
 		if p == noResolve {
+			return true
+		}
+	}
+	return false
+}
+
+func HasFullMatch(params []string) bool {
+	for _, p := range params {
+		if p == fullMatch {
 			return true
 		}
 	}
--- a/rules/parser.go
+++ b/rules/parser.go
@@ -32,7 +32,8 @@ func ParseRule(tp, payload, target string, params []string) (C.Rule, error) {
 	case "DST-PORT":
 		parsed, parseErr = NewPort(payload, target, false)
 	case "PROCESS-NAME":
-		parsed, parseErr = NewProcess(payload, target)
+		fullMatch := HasFullMatch(params)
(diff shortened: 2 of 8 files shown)


Kr328:
Thanks for your patch

But some process change its name by modify `cmdline`(or `argv[*]`)

#### Example
In Android, app can be identify by **package name**, so android system rename the process to `<package name>:<subprocess>`, it cause `exe` not respond process name

##### `cmdline`
```bash
# cat /proc/`pidof com.github.kr328.clash:background`/cmdline
com.github.kr328.clash:background
``` 

##### `exe`
```bash
# readlink /proc/`pidof com.github.kr328.clash:background`/exe
/system/bin/app_process64
```

icpz:
Thanks, I've reverted that part.

Dreamacro:
Is there any real-world case that needs to use `full-match`?

icpz:
> Is there any real-world case that needs to use `full-match`?

I don't known yet... This is just a possible use case.

Maybe someone want `/usr/lib/apt/methods/ftp` to go through specific proxy, and keep `/usr/bin/ftp` matching other rules in the meantime.

Dreamacro:
> Maybe someone want /usr/lib/apt/methods/ftp to go through specific proxy, and keep /usr/bin/ftp matching other rules in the meantime.

This is not convincing 🤔

icpz:
Closed for now. This PR might be reopened if somebody comes up with any meaningful case.

EKyero:
> Is there any real-world case that needs to use `full-match`?

在Windows上重名的情况挺多的,典型的便是Launcher.exe和Update.exe

PR #947, opened 2020-09-05
Fix: PROCESS-NAME on FreeBSD 11.x

icpz:
FreeBSD 11 and 12 have different PCB layouts. Now class will check the major version while initialization.


Dreamacro:
Has this change been tested?

icpz:
I've tested it on my virtual machines, including freebsd 10.x(on which disables the process-name rules automatically)/11.x/12.x

Dreamacro:
LGTM, Thanks for your hard work.

PR #954, opened 2020-09-06
Feature: Adds Plain TCP Redirect Outbound

mayanez:
Hi,

Cool project! I'm not sure what the proper contribution guidelines are so let me know if I should modify anything.

This add a simple TCP redirection outbound. This is the same functionality as the `redirect` in V2Ray's `freedom` [protocol](https://www.v2ray.com/en/configuration/protocols/freedom.html). For example, it can be used to redirect traffic to different gateways on a network.

The same functionality can be implemented using `sockopt` mark/tags, but since it differs depending on the Firewall it might be useful for this to be included as part of Clash. I have yet to measure the `sockopt` mark option's performance, but I suspect it might be better. Given that it might make sense to eventually continue with PR https://github.com/Dreamacro/clash/pull/162 and add the ability to mark each outbound with its own `SO_MARK`.

### DIFF
--- a/adapters/outbound/shadowsocks.go
+++ b/adapters/outbound/shadowsocks.go
@@ -38,6 +38,7 @@ type ShadowSocksOption struct {
 	Plugin     string                 `proxy:"plugin,omitempty"`
 	PluginOpts map[string]interface{} `proxy:"plugin-opts,omitempty"`
 	SocketMark string                 `proxy:"socket-mark,omitempty"`
+	Interface  string                 `proxy:"interface-name,omitempty"`
 }
 
 type simpleObfsOption struct {
@@ -75,7 +76,7 @@ func (ss *ShadowSocks) StreamConn(c net.Conn, metadata *C.Metadata) (net.Conn, e
 }
 
 func (ss *ShadowSocks) DialContext(ctx context.Context, metadata *C.Metadata) (C.Conn, error) {
-	c, err := dialer.DialContext(ctx, "tcp", ss.addr, ss.SocketMark())
+	c, err := dialer.DialContext(ctx, "tcp", ss.addr, dialer.DialOptions{SocketMark: ss.SocketMark(), Interface: ss.Interface()})
 	if err != nil {
 		return nil, fmt.Errorf("%s connect error: %w", ss.addr, err)
 	}
@@ -162,6 +163,7 @@ func NewShadowSocks(option ShadowSocksOption) (*ShadowSocks, error) {
 			tp:         C.Shadowsocks,
 			udp:        option.UDP,
 			socketmark: option.SocketMark,
+			ifname:     option.Interface,
 		},
 		cipher: ciph,
 
--- a/adapters/outbound/shadowsocks.go
+++ b/adapters/outbound/shadowsocks.go
@@ -37,6 +37,7 @@ type ShadowSocksOption struct {
 	UDP        bool                   `proxy:"udp,omitempty"`
 	Plugin     string                 `proxy:"plugin,omitempty"`
 	PluginOpts map[string]interface{} `proxy:"plugin-opts,omitempty"`
+	SocketMark string                 `proxy:"socket-mark,omitempty"`
 }
 
 type simpleObfsOption struct {
@@ -74,7 +75,7 @@ func (ss *ShadowSocks) StreamConn(c net.Conn, metadata *C.Metadata) (net.Conn, e
 }
 
 func (ss *ShadowSocks) DialContext(ctx context.Context, metadata *C.Metadata) (C.Conn, error) {
-	c, err := dialer.DialContext(ctx, "tcp", ss.addr)
+	c, err := dialer.DialContext(ctx, "tcp", ss.addr, ss.SocketMark())
 	if err != nil {
 		return nil, fmt.Errorf("%s connect error: %w", ss.addr, err)
 	}
@@ -156,10 +157,11 @@ func NewShadowSocks(option ShadowSocksOption) (*ShadowSocks, error) {
 
 	return &ShadowSocks{
 		Base: &Base{
-			name: option.Name,
-			addr: addr,
-			tp:   C.Shadowsocks,
-			udp:  option.UDP,
+			name:       option.Name,
+			addr:       addr,
+			tp:         C.Shadowsocks,
+			udp:        option.UDP,
+			socketmark: option.SocketMark,
 		},
 		cipher: ciph,
 
--- a/common/sockopt/sobind_darwin.go
+++ b/common/sockopt/sobind_darwin.go
@@ -1,0 +1,19 @@
+package sockopt
+
+import (
+	"github.com/Dreamacro/clash/log"
+	"net"
+	"syscall"
+)
+
+func BindToDevice(dialer *net.Dialer, ifname string) {
+	iface, err := net.InterfaceByName(ifname)
+
+	dialer.Control = func(network, address string, c syscall.RawConn) error {
+		return c.Control(func(fd uintptr) {
+			if err := syscall.SetsockoptInt(int(fd), syscall.IPPROTO_IP, syscall.IP_BOUND_IF, iface.index); err != nil {
+				log.Errorln("Sockopt IP_BOUND_IP error: %s", err)
+			}
+		})
+	}
+}
--- a/component/dialer/dialer.go
+++ b/component/dialer/dialer.go
@@ -9,6 +9,11 @@ import (
 	"strconv"
 )
 
+type DialOptions struct {
+	SocketMark string
+	Interface  string
+}
+
 func Dialer() (*net.Dialer, error) {
 	dialer := &net.Dialer{}
 	if DialerHook != nil {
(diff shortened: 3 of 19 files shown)


Dreamacro:
Can you elaborate on the specific usage scenarios of `redirect`? I don't understand which scenario this is used in.

mayanez:
Sure, I’ll start with my use case. I’m using it to redirect traffic to different gateways on my internal network. These gateways are connected to different VPN endpoints effectively allowing me to split tunnel traffic.

As for other use-cases, the redirect adds an additional layer of abstraction to push the tunneling of traffic to other nodes/programs (eg. Wireguard, OpenVPN, etc) with Clash doing rule based routing by URLs etc. Put simply, it makes Clash a SNI based router. What makes this setup super powerful is the useful lists available for Clash.

Dreamacro:
> As for other use-cases, the redirect adds an additional layer of abstraction to push the tunneling of traffic to other nodes/programs (eg. Wireguard, OpenVPN, etc) with Clash doing rule-based routing by URLs, etc.

The core of the proxy client is how it receives traffic (i.e., Inbound). There are several types of traffic that clash currently receives: HTTP, Socks5, Redir, and TUN (closed source implementation). There are some problems with Redir and TUN (i.e., forcing traffic through iptable).

When using these two types, you usually need to specify `interface-name`, and all `Proxy`s will be forced to make requests through this interface. The VPN implementation is via TUN (on macOS), so this type is likely to cause problems in both modes.

I'm not a native English speaker, so I hope you understand what I'm saying.

mayanez:
No worries, I think I followed. 

Correct me if I'm wrong, but this is my current understanding. 

The `interface-name` is useful (1) when you are running on the same host and (2) when you want to send **all** your outbound connections via the specific interface. That is, the `interface-name` feature does **not** allow you to route individual rules via different interfaces or across different machines.  For example, you cannot do this:
```
               eth0     +----------------+
         +------------->+ VPN Provider 1 |
    +---------+         +----------------+
    |  Clash  |
    +---------+         +----------------+
         +------------->+ VPN Provider 2 |
               eth1     +----------------+
```


The TCP redirect I'm proposing allows you to route rules individually allowing you what I described above via `iptables`. 

Additionally, the machine that is running `clash` does not need to be the same one that tunnels the traffic. For example, a separate machine with a different OS can be setup to transparently tunnel data through a VPN. 
```
url: google.com
      10.0.0.10    +---------+    +----------------+
    +------------->+ Gateway +----> VPN Provider 1 |
+-------+          +---------+    +----------------+
| Clash |
+-------+          +---------+    +----------------+
    +------------->+ Gateway +----> VPN Provider 2 |
    url: bing.com  +---------+    +----------------+
      10.0.0.20
```

Hopefully, the diagrams make the communication easier :smiley:

Dreamacro:
If what you want is the second diagram, it's has a question about your PR: the TCP redirect problem.

When you dial a TCP request, there are two tuples: (srcIP, srcPort) --> (dstIP, dstPort). In your implementation just dialing a TCP connection, the gateway is not clear to the receiver what you really want to request, and need a protocol to received (e.g. Socks5 or shadowsocks).

```
want to connect google.com
       +---------+                 +---------+
       |  clash  |  +----------->  | Gateway | Not sure where to connect(google.com missing).
       +---------+10.10.10.10:1234 +---------+

```

Dreamacro:
Let me say what I have met (which may not be the same but is very similar).
1. the company needs to install a complicated application to connect to the VPN (the company's domain is `*.example.com`)
2. install the application in the virtual machine (10.10.10.10) and connect to the VPN so that the virtual machine can connect to `*.example.com`
3. run a clash in the virtual machine
```yaml
# two line to start a socks5 server
socks-port: 7890
allow-lan: true
```
4. run another clash on host.
```yaml
proxies:
    - name: socks5
      server: 10.10.10.10
      port: 7890
rules:
  - DOMAIN-SUFFIX,example.com,socks5
```

mayanez:
I think you're totally right for the second scenario as you wouldn't be able to NAT traffic back.

However, you can still do the first scenario. Actually, it's simpler to use `SO_MARK`. Therefore I've gone ahead and added that. Let me know if the implementation looks good. I am still trying to learn the codebase and go at the same time.

You can set the `socket-mark` globally in the config so it applies to `DIRECT`. 

You can also add `socket-mark` to each proxy, for example:
```
proxies:
  - name: "ssr"
    type: ssr
    socket-mark: 10
```

You can also now create multiple "direct" type proxies which is the same as "DIRECT", but you can define them with different `socket-mark`.
```
proxies:
  - name: "direct10"
    type: direct
    socket-mark: 10
```

With the `SO_MARK` then you can freely route traffic to different interfaces etc.

 Any thoughts?

Dreamacro:
But SO_MARK only works on Linux. macOS and Windows don't support it. I think `interface-name` is the best solution. Now `interface-name` is implemented via local-address, but I'm going to use syscall for macOS and Linux.

mayanez:
I think it's beneficial to add `SO_MARK` support even if it's just for Linux as it adds flexibility when it comes to writing `iptables` rules. 

For multi-platform support, I agree that extending `interface-name` per proxy is a good solution.

mayanez:
I went ahead and added the `interface-name` support for both Linux and macOS. I don't have a Mac to verify, but it seems to be the right system call. 

Let me know if are able to verify it.

mayanez:
> If what you want is the second diagram, it's has a question about your PR: the TCP redirect problem.
> 
> When you dial a TCP request, there are two tuples: (srcIP, srcPort) --> (dstIP, dstPort). In your implementation just dialing a TCP connection, the gateway is not clear to the receiver what you really want to request, and need a protocol to received (e.g. Socks5 or shadowsocks).
> 
> ```
> want to connect google.com
>        +---------+                 +---------+
>        |  clash  |  +----------->  | Gateway | Not sure where to connect(google.com missing).
>        +---------+10.10.10.10:1234 +---------+
> ```

After some research, it seems there is a solution to support this scenario (see the [HAProxy Protocol RFC](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt)). Not sure if it's something you'd eventually be interested in, but I will add it here for future reference.

adapters/outbound/shadowsocks.go
```diff
@@ -74,7 +76,7 @@ func (ss *ShadowSocks) StreamConn(c net.Conn, metadata *C.Metadata) (net.Conn, e
 }
 
 func (ss *ShadowSocks) DialContext(ctx context.Context, metadata *C.Metadata) (C.Conn, error) {
-	c, err := dialer.DialContext(ctx, "tcp", ss.addr)
+	c, err := dialer.DialContext(ctx, "tcp", ss.addr, dialer.DialOptions{SocketMark: ss.SocketMark(), Interface: ss.Interface()})
```
Dreamacro, line 79: It seems that you're the only handling with the TCP-related parts, but there are some proxies that contain UDP parts (shadowsocks, socks5, etc.)

common/sockopt/sobind_darwin.go
```diff
@@ -0,0 +1,19 @@
+package sockopt
+
+import (
+	"github.com/Dreamacro/clash/log"
```
Dreamacro, line 4: In this project, import split 3 parts: std, internal and third party, so it's better to format with

```go
import (
  // std

  // internal

  // third party
)
```

common/sockopt/sobind_darwin.go
```diff
@@ -0,0 +1,19 @@
...
+	"syscall"
+)
+
+func BindToDevice(dialer *net.Dialer, ifname string) {
+	iface, err := net.InterfaceByName(ifname)
+
+	dialer.Control = func(network, address string, c syscall.RawConn) error {
+		return c.Control(func(fd uintptr) {
+			if err := syscall.SetsockoptInt(int(fd), syscall.IPPROTO_IP, syscall.IP_BOUND_IF, iface.index); err != nil {
+				log.Errorln("Sockopt IP_BOUND_IP error: %s", err)
```
Dreamacro, line 15: Previously, clash logging followed these rules. If it affects clash itself, the log level is `error`; if it doesn't affect clash itself, but there is an error, the log level is `warning`.

component/dialer/dialer.go
```diff
@@ -125,6 +141,14 @@ func dualStackDialContext(ctx context.Context, network, address string) (net.Con
 			return
 		}
 
+		if opts != nil {
+			markint, err := strconv.Atoi(opts[0].SocketMark)
```
Dreamacro, line 145: what if `DialOptions` has two options?

Dreamacro (changes requested):
Sorry for not checking this pr in a while, it's been a busy time. Here are a few of the tiny problems I found

adapters/outbound/shadowsocks.go
mayanez: Got it. I will look into the UDP parts as well.

component/dialer/dialer.go
mayanez: This can definitely be changed to account for an arbitrary amount of options.

Dreamacro:
In order to make as small a change as possible (compared to this PR), I'm going to write a commit first to add the Linux and macOS syscall binding to the interface.

Dreamacro:
Sorry for the long time without reply, now I have enough time and energy to complete this feature request. I want to know if you still have the intention to complete this pr? If not, I'll finish this pr. I'll rebase to the latest code first

PR #964, opened 2020-09-13
Feature: support domain in fallback filter

MelvinTo:
Sometimes domain (e.g. google.com) is resolved to a random IP address in China, the current clash will return the IP from standard nameservers directly, which is not right.

This new config is to always return the result of fallback nameservers if domain or domain suffix is matched.

Fallback Config Example:

```
  fallback-filter:
    geoip: true
    ipcidr:
      - 240.0.0.0/4
      - 0.0.0.0/32
    domain:
      - '+.google.com'
      - '+.facebook.com'
      - '+.youtube.com'
```

### DIFF
--- a/dns/filters.go
+++ b/dns/filters.go
@@ -4,9 +4,10 @@ import (
 	"net"
 
 	"github.com/Dreamacro/clash/component/mmdb"
+	"github.com/Dreamacro/clash/component/trie"
 )
 
-type fallbackFilter interface {
+type fallbackIPFilter interface {
 	Match(net.IP) bool
 }
 
@@ -23,4 +24,23 @@ type ipnetFilter struct {
 
 func (inf *ipnetFilter) Match(ip net.IP) bool {
 	return inf.ipnet.Contains(ip)
+}
+
+type fallbackDomainFilter interface {
+	Match(domain string) bool
+}
+type domainFilter struct {
+	tree *trie.DomainTrie
+}
+
+func NewDomainFilter(domains []string) *domainFilter {
+	df := domainFilter{tree: trie.New()}
+	for _, domain := range domains {
+		df.tree.Insert(domain, "")
+	}
+	return &df
+}
+
+func (df *domainFilter) Match(domain string) bool {
+	return df.tree.Search(domain) != nil
 }
--- a/dns/resolver.go
+++ b/dns/resolver.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math/rand"
 	"net"
+	"strings"
 	"time"
 
 	"github.com/Dreamacro/clash/common/cache"
@@ -34,13 +35,14 @@ type result struct {
 }
 
 type Resolver struct {
-	ipv6            bool
-	hosts           *trie.DomainTrie
-	main            []dnsClient
-	fallback        []dnsClient
-	fallbackFilters []fallbackFilter
-	group           singleflight.Group
-	lruCache        *cache.LruCache
+	ipv6                  bool
+	hosts                 *trie.DomainTrie
+	main                  []dnsClient
+	fallback              []dnsClient
+	fallbackDomainFilters []fallbackDomainFilter
+	fallbackIPFilters     []fallbackIPFilter
+	group                 singleflight.Group
+	lruCache              *cache.LruCache
 }
 
 // ResolveIP request with TypeA and TypeAAAA, priority return TypeA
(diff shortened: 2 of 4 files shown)


Dreamacro:
But the domain match timing is different from geoip and ipcidr. It should be resolved by fallback only

MelvinTo:
Agree. Let me check code

MelvinTo:
code updated

dns/filters.go
```diff
@@ -2,25 +2,51 @@ package dns
...
-func (inf *ipnetFilter) Match(ip net.IP) bool {
+func (inf *ipnetFilter) MatchIP(ip net.IP) bool {
 	return inf.ipnet.Contains(ip)
 }
+
+func (inf *ipnetFilter) MatchDomain(domain string) bool {
+	return false
+}
+
+type domainFilter struct {
```
Dreamacro, line 38: Why not use "github.com/Dreamacro/clash/component/trie"?

dns/resolver.go
```diff
@@ -214,26 +224,52 @@ func (r *Resolver) batchExchange(clients []dnsClient, m *D.Msg) (msg *D.Msg, err
...
+		for _, domain := range domains {
+			if r.shouldFallbackExclusively(domain) {
+				return false
 			}
 		}
 	}
+	return true
+}
+
+func (r *Resolver) fallbackExchange(m *D.Msg) (msg *D.Msg, err error) {
```
Dreamacro, line 242: maybe change this function name to `ipExchange`, then adding a `shouldQueryFallbackOnly` logic seems better?

dns/resolver.go
```diff
@@ -284,6 +320,16 @@ func (r *Resolver) msgToIP(msg *D.Msg) []net.IP {
 	return ips
 }
 
+func (r *Resolver) msgToDomains(msg *D.Msg) []string {
+	domains := []string{}
+
+	for _, question := range msg.Question {
```
Dreamacro, line 326: just handle msg.Question[0] is enough

Dreamacro (commented):
sorry for the late, here are some of my thoughts.

MelvinTo:
updated, please review

MelvinTo:
one issue is how to combine *.google.com and google.com with trie? now the config looks redundant.

Dreamacro:
> one issue is how to combine *.google.com and google.com with trie? now the config looks redundant.

`+.google.com`

MelvinTo:
> `+.google.com`

cool, example updated.

dns/resolver.go
```diff
@@ -286,14 +327,17 @@ func NewResolver(config Config) *Resolver {
...
 	}
 	for _, ipnet := range config.FallbackFilter.IPCIDR {
-		fallbackFilters = append(fallbackFilters, &ipnetFilter{ipnet: ipnet})
+		fallbackIPFilters = append(fallbackIPFilters, &ipnetFilter{ipnet: ipnet})
 	}
-	r.fallbackFilters = fallbackFilters
+	r.fallbackIPFilters = fallbackIPFilters
+
+	fallbackDomainFilters := []fallbackDomainFilter{NewDomainFilter(config.FallbackFilter.Domain)}
+	r.fallbackDomainFilters = fallbackDomainFilters
```
Dreamacro, line 340: maybe we can leave `r.fallbackDomainFilters` nil when `len(config.FallbackFilter.Domain) == 0`

dns/resolver.go
```diff
@@ -170,19 +172,49 @@ func (r *Resolver) batchExchange(clients []dnsClient, m *D.Msg) (msg *D.Msg, err
 	return
 }
 
-func (r *Resolver) fallbackExchange(m *D.Msg) (msg *D.Msg, err error) {
-	msgCh := r.asyncExchange(r.main, m)
+func (r *Resolver) shouldOnlyQueryFallback(m *D.Msg) bool {
 	if r.fallback == nil {
```
Dreamacro, line 176: if r.fallback == nil || len(r.fallbackDomainFilters) == 0 {}

Dreamacro (changes requested):

Loyalsoldier:
I think it's better if support `rule-providers`.

Dreamacro:
@Loyalsoldier Not sure what the relationship between this pr and rule provider.

Dreamacro:
@MelvinTo Thanks for your hard work

Loyalsoldier:
> @Loyalsoldier Not sure what the relationship between this pr and rule provider.

I mean support `rule-providers` files in `fallback-filter` section so as not to specify domains and CIDR manually, which is inconvenient.

Dreamacro:
@Loyalsoldier The need for an external file or an external link depends on the number of ipcidr rules. IMO, fallback rules shouldn't be large, so maybe we should not spend too much time doing this.

Loyalsoldier:
Or to support an option like `force-remote-dns` in Surge to skip the DNS resolving process for domain type of rules and rule-set.

PR #1020, opened 2020-10-16
Feature: reload config on SIGHUP

blip:
Currently, to reload the config without restarting clash, the controller API is the only choice, which is not init friendly.

This PR makes clash reload config on SIGHUP. It has been manually tested.

On an unrelated note, I'd like to make two more contributions:

1. Make clash call `ifup`/`ifdown` script when it brings up/tears down the tun interface in tun mode.
2. Make clash reject unmapped fake IPs (IPs in the fake IP range not yet being leased from the DNS) in tun mode.

`#1` is, again, to make clash more init friendly. The routing table can only be manipulated once the tun interface is up, and there is currently no easy way to wait for it other than polling.
`#2` is to prevent the infinite loop where such packets get routed back to the tun interface after they go out. I don't think it's possible to break the loop via packet filtering, since clash does establish legit connections within the fake IP range, and there is no easy way to discern the two cases from the outside.

These contributions only pertain to the closed-source premium version. I wonder if opening feature requests is the only way to expedite their implementations?

### DIFF
--- a/main.go
+++ b/main.go
@@ -97,6 +97,20 @@ func main() {
 		log.Fatalln("Parse config error: %s", err.Error())
 	}
 
+	go func() {
+		sigCh := make(chan os.Signal, 1)
+		signal.Notify(sigCh, syscall.SIGHUP)
+		for range sigCh {
+			log.Infoln("SIGHUP received, reloading config")
+			cfg, err := executor.Parse()
+			if err != nil {
+				log.Warnln("Parse config error: %s", err.Error())
+				continue
+			}
+			executor.ApplyConfig(cfg, true)
+		}
+	}()
+
 	sigCh := make(chan os.Signal, 1)
 	signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
 	<-sigCh


Dreamacro:
There is a problem here, if you have used the HTTP API to load a configuration for another path, then the default config path (~/.config/clash/config.yaml) will be reloaded. so in some cases, it may not work as expected.

blip:
Do you mean [this case](https://github.com/Dreamacro/clash/blob/d3b14c325f4487f57a3e2e8c7dca73e1a937ccdc/hub/route/configs.go#L113)?

It currently doesn't save the specified path. Should that be changed?

blip:
Actually, I’m not sure if it’s worth the effort. It’s very unlikely people who rely on an init system to reload the config would want to dynamical switch to a different config file via the API (the official dashboard doesn’t offer this feature anywhere if I’m not wrong) and vice versa.

And in the unlikely event that some people do perform this weird combination, it shouldn’t cause any trouble.

This is not an edge case that is worth optimize for imho.

Dreamacro:
It's just that without any description, this feature makes it confusing (whether to load the default configuration or the latest one)

> 1. Make clash call ifup/ifdown script when it brings up/tears down the tun interface in tun mode.
just call API with

```yaml
tun:
    enable: true # false would close tun interface
```

> 2. Make clash reject unmapped fake IPs (IPs in the fake IP range not yet being leased from the DNS) in tun mode.
In fact, it is that way now.

https://github.com/Dreamacro/clash/blob/master/tunnel/tunnel.go#L137-L139

blip:
> whether to load the default configuration or the latest one

How about we document it and say it always loads the default configuration? This should cover most, if not all, use cases. 

> just call API with

What API? The code you pasted seems to be clash config, and it only enables the tun mode.

> In fact, it is that way now.

Does the tun mode also use this logic? I can reliably reproduce in clash 2020.10.09 by executing `curl 192.18.54.1` in a LAN machine and see the following logs quickly populate:
```
[info] [TCP] 198.18.0.1:44320 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44322 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44324 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44326 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44328 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44330 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44332 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44334 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44336 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44338 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44340 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44342 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44344 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44346 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44348 --> 198.18.54.1 using DIRECT by Script
[info] [TCP] 198.18.0.1:44350 --> 198.18.54.1 using DIRECT by Script
```

Dreamacro:
> Does the tun mode also use this logic? I can reliably reproduce in clash Premium 2020.10.09 by executing curl 192.18.54.1 in a LAN machine and see the following logs quickly populate:

This was a bug that was generated by cherry-pick after #936 and has now been fixed. If you use docker, you can just use dreamacro/clash-premium

blip:
Thanks for the quick fix. Just confirmed it.

PR #1044, opened 2020-10-28
Feature: support round-robin strategy for load-balance group

uchuhimo:
Add support for round-robin strategy for load-balance group.
Using this strategy, the requests will be distributed evenly to all the proxies.
This feature makes clash suitable to act as a load balancer for a web crawler.

## Design

`load-balance` group will have two strategies:

- `consistent-hashing` strategy: the default strategy, which behaves exactly the same as the original `load-balance` group
- `round-robin` strategy: the new strategy, which behaves as described above

You can use the `consistent-hashing` strategy as follow:

```yaml
  - name: "load-balance"
    type: load-balance
```

or

```yaml
  - name: "load-balance"
    type: load-balance
    strategy: consistent-hashing
```

You can use the `round-robin` strategy as follow:

```yaml
  - name: "load-balance"
    type: load-balance
    strategy: round-robin
```


Dreamacro:
I think it would be better to provide a `strategy` parameter in the `load-balance'. And `Consistent Hashing` is the default

uchuhimo:
OK, so the original `load-balance` will be:

```yaml
  - name: "load-balance"
    type: load-balance
```

or

```yaml
  - name: "load-balance"
    type: load-balance
    strategy: consistent-hashing
```

The round-robin strategy will be:

```yaml
  - name: "load-balance"
    type: load-balance
    strategy: round-robin
```

right?

Dreamacro:
@uchuhimo okay, that's what I thought.

uchuhimo:
@Dreamacro I have updated both the description and the code. The updated code has been checked on my crawler. Please check it.

adapters/outboundgroup/roundrobin.go
```diff
@@ -0,0 +1,89 @@
+package outboundgroup
```
Dreamacro, line 1: Probably shouldn't create a new group because only one function in the original `load-balance` was inconsistent.

https://github.com/Dreamacro/clash/blob/b1795b1e3d13a2c8898fc457b7b8b4147b41f4ee/adapters/outboundgroup/loadbalance.go#L84-L97

Dreamacro (changes requested):

adapters/outboundgroup/roundrobin.go
uchuhimo: I need your advice since I'm not familiar with Golang (I have never used Golang before). If I use a single group `LoadBalance` for these two strategies, then there are two ways to model them:

- use condition branch in `Unwrap`, which is simple but brings runtime overhead
- let two strategies be two subclasses `ConsistentHashing` and `RoundRobin` of a base class `LoadBalance`

Which would you like to use? Any better advice?

uchuhimo (commented):
@Dreamacro  I try to use the subclass solution. Any advice is welcome.

Dreamacro:
Using a strategy func would be simpler
```go
type strategyFn = func(proxies []C.Proxy, metadata *C.Metadata) C.Proxy

func strategyRoundRobin() strategyFn {
    idx := 0
    return func(proxies []C.Proxy, metadata *C.Metadata) C.Proxy {
       length := len(proxies)
       for i := 0; i < length; i++ {
		idx = (idx + 1) % length
		proxy := proxies[idx]
		if proxy.Alive() {
			return proxy
		}
	}

	return proxies[0]
    }
}

func strategyConsistentHashing() strategyFn {
    maxRetry := 5
    return func(proxies []C.Proxy, metadata *C.Metadata) C.Proxy {
        key := uint64(murmur3.Sum32([]byte(getKey(metadata))))
        buckets := int32(len(proxies))
        for i := 0; i < maxRetry; i, key = i+1, key+1 {
		idx := jumpHash(key, buckets)
		proxy := proxies[idx]
		if proxy.Alive() {
			return proxy
		}
	}

	return proxies[0]
    }
}

func (lb *LoadBalance) Unwrap(metadata *C.Metadata) C.Proxy {
	proxies := lb.proxies()
	return lb.strategyFn(proxies, metadata)
}

```

uchuhimo:
@Dreamacro Got it. I have updated the code.

adapters/outboundgroup/parser.go
```diff
@@ -20,6 +21,7 @@ var (
 type GroupCommonOption struct {
 	Name     string   `group:"name"`
 	Type     string   `group:"type"`
+	Strategy string   `group:"strategy,omitempty"`
```
Dreamacro, line 24: `Strategy` is not a common option, you can checkout `parseURLTestOption`

uchuhimo: I have moved it into `loadbalance.go`.

adapters/outboundgroup/loadbalance.go
```diff
@@ -23,6 +25,20 @@ type LoadBalance struct {
 	strategyFn strategyFn
 }
 
+var (
+	errMissStrategy = errors.New("`strategy` missing")
+	errStrategy     = errors.New("unsupported strategy")
+	)
```
Dreamacro, line 31: `go fmt ./...`

adapters/outboundgroup/loadbalance.go
```diff
@@ -23,6 +25,20 @@ type LoadBalance struct {
...
+
+func parseStrategy(config map[string]interface{}) (strategy string, err error) {
+	if elm, ok := config["strategy"]; ok {
+		if strategy, ok := elm.(string); ok {
+			return strategy, nil
+		}
+	}
+	return "", errMissStrategy
```
Dreamacro, line 39: `consistent-hashing` is the default value

Dreamacro: maybe this function return strategyFn directly

Dreamacro (changes requested):

Dreamacro:
Thanks for your contribution

uchuhimo:
@Dreamacro It seems that my last commit is not merged. Should I open a new PR?

Dreamacro:
@uchuhimo Currently this is a bit simpler in parser.go, It's okay to keep it.

uchuhimo:
Got it.

PR #1048, opened 2020-10-29
Fix: tunnel manager & tracker race condition

xjasonlyu:
fix https://github.com/Dreamacro/clash/issues/1041 by using `go.uber.org/atomic`

### DIFF
--- a/adapters/outbound/base.go
+++ b/adapters/outbound/base.go
@@ -160,11 +160,7 @@ func (p *Proxy) MarshalJSON() ([]byte, error) {
 // URLTest get the delay for the specified URL
 func (p *Proxy) URLTest(ctx context.Context, url string) (t uint16, err error) {
 	defer func() {
-		if err == nil {
-			p.alive.Store(true)
-		} else {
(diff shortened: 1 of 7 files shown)


Dreamacro:
To use `go.uber.org/atomic`, you can replace all atomic with the new `atomic`

xjasonlyu:
> To use `go.uber.org/atomic`, you can replace all atomic with the new `atomic`

Done👌

adapters/outbound/base.go
```diff
@@ -160,9 +161,9 @@ func (p *Proxy) MarshalJSON() ([]byte, error) {
 func (p *Proxy) URLTest(ctx context.Context, url string) (t uint16, err error) {
 	defer func() {
 		if err == nil {
```
Dreamacro, line 163: p.alive.Store(err == nil) 🤣

xjasonlyu: 光改代码忘记看逻辑了🌚

Dreamacro:
LGTM, thanks

PR #1140, opened 2020-12-22
Fix: PROCESS-NAME rule for UDP sessions on Windows

icpz:
Windows binds an unbound UDP socket to 0.0.0.0/:: while first `sendto`.

So we give unspecific addresses a try for UDP sessions.


Dreamacro:
LGTM, Thanks 👍

PR #1191, opened 2021-01-20
fix docker label

junjieyuan:
Signed-off-by: Junjie Yuan <yuan@junjie.pro>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN go mod download && \
     mv ./bin/clash-docker /clash
 
 FROM alpine:latest
-LABEL org.opencontainers.image.source https://github.com/Dreamacro/clash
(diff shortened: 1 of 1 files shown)


junjieyuan:
See [docker docs](https://docs.docker.com/engine/reference/builder/#label).

Dreamacro:
`LABEL a=foo` and `LABEL a foo` should be equivalent, how to prove it:

```
$ docker pull dreamacro/clash
$ docker inspect dreamacro/clash
```

junjieyuan:
Yes, they are equal. But the docker docs did not state that this is the correct way to use it, and GoLand marked it as an error. Should we modify it?

GoLand screenshot

Dreamacro:
Okay, Standardize it.

Dreamacro:
Thanks

PR #1319, opened 2021-04-01
[New Feature] Use mixed-port instead of port

CoelWu:
Since ``mixed-port`` support both ``socks5`` and ``http`` inbound, there is no need for a initial config to accept ``socks5`` only


Dreamacro:
LGTM, thanks

PR #1351, opened 2021-04-14
Chore: PROCESS-NAME support freebsd 13, fix panic on unsupported platforms

icpz:
FreeBSD 13 has been released: [https://www.freebsd.org/releases/13.0R/announce](https://www.freebsd.org/releases/13.0R/announce).

Luckily the offsets stay same as 12.x

Tested on FreeBSD 12.2 & FreeBSD 13.


Dreamacro:
LGTM, Thanks!

PR #1406, opened 2021-05-16
Feature: lookup domain through assigned nameserver

icpz:
Sometimes we need clash lookup domains via specific nameservers. Here is an example configuration:
```yaml
dns:
  enable: true
  listen: '0.0.0.0:53'
  assign-nameserver:
    'www.baidu.com': '114.114.114.114'
    '+.internal.crop.com': '10.0.0.1'
```

This feature may resolve issues like https://github.com/Dreamacro/clash/issues/1376 .

It seems that we need a more appropriated term here, instead of `assign-nameserver`.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -64,6 +64,7 @@ type DNS struct {
 	DefaultNameserver []dns.NameServer `yaml:"default-nameserver"`
 	FakeIPRange       *fakeip.Pool
 	Hosts             *trie.DomainTrie
+	AssignNameServer  map[string]dns.NameServer
 }
 
 // FallbackFilter config
@@ -106,6 +107,7 @@ type RawDNS struct {
 	FakeIPRange       string            `yaml:"fake-ip-range"`
 	FakeIPFilter      []string          `yaml:"fake-ip-filter"`
 	DefaultNameserver []string          `yaml:"default-nameserver"`
+	AssignNameServer  map[string]string `yaml:"assign-nameserver"`
 }
 
 type RawFallbackFilter struct {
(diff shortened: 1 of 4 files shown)


icpz:
Now the config looks like:
```yaml
dns:
  resolver-rule:
    enable: true
    listen: '0.0.0.0:53'
    assign-nameserver:
      'www.baidu.com': '114.114.114.114'
      '+.internal.crop.com': '10.0.0.1'
```

config/config.go
```diff
@@ -500,6 +502,22 @@ func parseNameServer(servers []string) ([]dns.NameServer, error) {
 	return nameservers, nil
 }
 
+func parseResolverRule(cfg *RawDNS) (map[string]dns.NameServer, error) {
+	rule := make(map[string]dns.NameServer)
```
Dreamacro, line 506: prefer `map[string]dns.NameServer{}`

config/config.go
```diff
@@ -500,6 +502,22 @@ func parseNameServer(servers []string) ([]dns.NameServer, error) {
...
+
+	if len(cfg.ResolverRule) != 0 {
+		for domain, server := range cfg.ResolverRule {
+			nameservers, err := parseNameServer([]string{server})
+			if err != nil {
+				return nil, err
+			}
+			rule[domain] = nameservers[0]
```
Dreamacro, line 514: The domain is not verified here, which make config unexpected

config/config.go
```diff
@@ -537,6 +555,10 @@ func parseDNS(cfg RawDNS, hosts *trie.DomainTrie) (*DNS, error) {
 		return nil, err
 	}
 
+	if dnsCfg.ResolverRule, err = parseResolverRule(&cfg); err != nil {
```
Dreamacro, line 558: should be just pass through a map?

dns/resolver.go
```diff
@@ -172,6 +177,20 @@ func (r *Resolver) batchExchange(clients []dnsClient, m *D.Msg) (msg *D.Msg, err
 	return
 }
 
+func (r *Resolver) matchResolverRule(m *D.Msg) []dnsClient {
+	domain := r.msgToDomain(m)
+	if domain == "" {
+		return []dnsClient{}
```
Dreamacro, line 183: `return nil` is enough

dns/resolver.go
```diff
@@ -172,6 +177,20 @@ func (r *Resolver) batchExchange(clients []dnsClient, m *D.Msg) (msg *D.Msg, err
...
+	}
+
+	record := r.rule.Search(strings.TrimRight(domain, "."))
```
Dreamacro, line 186: `msgToDomain` already trim right question

Dreamacro (changes requested):

config/config.go
icpz: Sorry I didn't get this... Could you please give an example?

Dreamacro (approved):

PR #1473, opened 2021-07-03
[Feature] support Xray's 0rtt ws

wwqgtxx:
fix broken https://github.com/Dreamacro/clash/pull/1298
smaller change compare with origin pull request and can work with v2ray-plugin without addition change

PR #1474, opened 2021-07-04
Fix the bug in handling `ws-path` 

monsoon235:
Fix the bug that clash can't correctly handle `ws-path` containing parameters (which used websocket early data in v2ray/xray)

### DIFF
--- a/transport/vmess/websocket.go
+++ b/transport/vmess/websocket.go
@@ -139,10 +139,15 @@ func StreamWebsocketConn(conn net.Conn, c *WebsocketConfig) (net.Conn, error) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Dreamacro:
The current clash has no `websocket early data`. and you are parsing the querystring in the wrong way

monsoon19260817:
> The current clash has no `websocket early data`. and you are parsing the querystring in the wrong way

I just remove the querystring in `ws-path`, which will be ignored when creating websocket connections, otherwise, `ws-path: /path?ed=2048` in config will leading to the bug which try to create websocket connections with `/path%3Fed=2048`

#1473 add real support for websocket early data. if #1473 will be merged, this pr can be ignored

Dreamacro:
same as #1558 with correct compatibility

PR #1480, opened 2021-07-07
Feature: support multiport condition for rules SRC-PORT and DST-PORT

yaling888:
Compatible with the old version
```yaml
rules:
  - SRC-PORT,7777,DIRECT
  - DST-PORT,123/136/137-139,DIRECT
```


Dreamacro:
What are your use-cases and why do you need to expand on PORT rules?

PR #1486, opened 2021-07-10
Add process name into metadata.

thatdms:
This is related to discusion https://github.com/Dreamacro/clash/discussions/1485.

The change will query process name for every connections.
This is slightly different from old behavior: only query process name when the program need to decide whether a connection match a PROCESS-NAME rule.

The old behavior is reasonable, if we assume the process name is only needed for matching the rule.
However, the information of process name could be useful in many other circumstances.
Scenario 1: A user wish to check the name of program that she is using, so as to set corresponding rule.
This might not be straightforward because: 1. the program name is not always same as application name shown in Task Manager. 2. An application could delegate the network function to certain other program, so the rule based on original program might not work.
Scenario 2: A user wish to check how many connection comes out of an application. 
This might deviate from the original goal of clash. However, this indeed requires the capability to capture the traffic of an application (even when it doesn't provide a native proxy support), and clash gracefully provide this functionality.

All above applications require collaboration from other software. I will show an example leveraging this PR in ClashX in a short time.

### DIFF
--- a/adapter/inbound/http.go
+++ b/adapter/inbound/http.go
@@ -17,5 +17,6 @@ func NewHTTP(target string, source net.Addr, conn net.Conn) *context.ConnContext
 		metadata.SrcIP = ip
 		metadata.SrcPort = port
 	}
+	metadata.ReadProcessName()
 	return context.NewConnContext(conn, metadata)
 }
--- a/adapter/inbound/https.go
+++ b/adapter/inbound/https.go
@@ -16,5 +16,6 @@ func NewHTTPS(request *http.Request, conn net.Conn) *context.ConnContext {
 		metadata.SrcIP = ip
 		metadata.SrcPort = port
 	}
+	metadata.ReadProcessName()
 	return context.NewConnContext(conn, metadata)
 }
--- a/adapter/inbound/packet.go
+++ b/adapter/inbound/packet.go
@@ -26,6 +26,7 @@ func NewPacket(target socks5.Addr, packet C.UDPPacket, source C.Type) *PacketAda
 		metadata.SrcPort = port
 	}
 
+	metadata.ReadProcessName()
 	return &PacketAdapter{
 		UDPPacket: packet,
 		metadata:  metadata,
--- a/adapter/inbound/socket.go
+++ b/adapter/inbound/socket.go
@@ -18,5 +18,6 @@ func NewSocket(target socks5.Addr, conn net.Conn, source C.Type) *context.ConnCo
 		metadata.SrcPort = port
 	}
 
+	metadata.ReadProcessName()
 	return context.NewConnContext(conn, metadata)
 }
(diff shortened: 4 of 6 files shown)


thatdms:
@Dreamacro 
It is only after I send the PR that I realize my modification is on an incomplete version of clash.
I feel frustrated because after one day of learning and writing, I still need to use the old software.
Could you please add me as collaborators of clash premium? I can integrate the function myself and therefore don't have to wait for merging of PR and next release.

Dreamacro:
The cost of "getting process name" is not the same on different OS, but it is also expensive.

So it's not a good idea to force process fetching at the inbound. In addition, when clash is used as a gateway, the process name is not available for connections from the remote IP.

I think a better solution would be to solve this problem at the GUI, or clash exposing an API for querying the connection process (e.g. as a parameter to the connections API). The cost is small enough to search or scan for all connections at once

Dreamacro:
I'm sorry I can't add you to the premium repository, I'd love to find someone to maintain the premium version together, but there are prerequisites:

* if the github account has good open-source behavior, no problem.
* if there is a lot of contribution to clash, no problem too.

thatdms:
Every problem is also an oppotunity. New function naturally introduce new design choices, together with different tradeoffs. However, given that no alternative design implemented currently, it is hard to precede the discussion. 
We don't even have a standard test of how much extra computation is introduced for resolving the process name. (Although my intuition is that the overhead is negligible, given the builtin cache and lower frequency of new connection, I cannot or don't have time to prove that.)

thatdms:
There is concerns on linking this account with other activities of mine. Therefore I don't believe I can present a CV for you to examine.
I am also not interested in being the full time developer of this software. I just want software that meet my own need (at the same time it is very likely to meet the needs of others). If my effort doesn't pay off (as I explained in last comment), or the work load needed is unclear (since it's not clear how many contribution and what specific contribution do you want), I guess I just have to find another way out.

Dreamacro:
A conservative estimate is that the logic of forcing process names to match is 10 to 20 times slower in my case

image

thatdms:
Thank you for the reply.

I am not sure how to interpret the value of the y-axis, the meaning of the spike in the figure, and how the multipliers are derived. 
If the figure should be interpreted as: it takes 50 microseconds (which is several orders of magnitude faster than clicking the mouse) to match the rules on new connection, then it seems to be imperceptible, at least for my daily use. Unfortunately, it seems that I won't be able to use it, as explained above.

BTW, are you willing to share the benchmark code? And does this benchmark work on open source version of clash?

Dreamacro:
it's grafana of https://github.com/Dreamacro/clash-tracing, using premium tracing API

```yaml
profile:
  # open tracing exporter API
  tracing: true
```

If you have and put PROCESS-NAME in the first line of the `rules`, the result should be similar to your PR result.

BTW, I deploy clash on macOS NUC, and DHCP server use it as my gateway. I have used less than 20 rules (Of course there is no `PROCESS-NAME`)

PR #1502, opened 2021-07-22
Fix: remove Content-Length from http CONNECT response

duament:
[rfc7231 section-4.3.6](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.6) states:

>   A server MUST NOT send any Transfer-Encoding or Content-Length header
>   fields in a 2xx (Successful) response to CONNECT.

In my case, `Content-Length` in response header breaks `ffplay`.

### Without the patch
```
CONNECT media.w3.org:443 HTTP/1.1
Host: 127.0.0.1:7890
Connection: close

HTTP/1.1 200 Connection established
Content-Length: 0
```

### With the patch
```
CONNECT media.w3.org:443 HTTP/1.1
Host: 127.0.0.1:7890
Connection: close

HTTP/1.1 200 Connection established
Connection: close
```


Dreamacro:
Thanks for your contribution 👏

Issue #1552, opened 2021-08-19
[BUG] panic: runtime error: invalid memory address or nil pointer dereference


Molay:
### Verify steps

- [ ]  如果你可以自己 debug 并解决的话，提交 PR 吧 Is this something you can **debug and fix**? Send a pull request! Bug fixes and documentation fixes are welcome. 
- [ ]  我已经在 [Issue Tracker](……/) 中找过我要提出的问题 I have searched on the [issue tracker](……/) for a related issue. 
- [ ]  我已经使用 dev 分支版本测试过，问题依旧存在 I have tested using the dev branch, and the issue still exists. 
- [X]  我已经仔细看过 [Documentation](https://github.com/Dreamacro/clash/wiki/) 并无法自行解决问题 I have read the [documentation](https://github.com/Dreamacro/clash/wiki/) and was unable to solve the issue. 
- [X]  这是 Clash 核心的问题，并非我所使用的 Clash 衍生版本（如 OpenClash、KoolClash 等）的特定问题 This is an issue of the Clash core *per se*, not to the derivatives of Clash, like OpenClash or KoolClash. 

### Clash version

higher than 2021.07.03

### What OS are you seeing the problem on?

Linux

### Clash config

```yaml
with general config
```

### Clash log

```shell
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xb8 pc=0x5a1d18]

goroutine 354 [running]:
github.com/Dreamacro/clash/listener/http.HandleConn({0x83e630, 0x40000a6708}, 0x40000a86c0, 0x40008c41e0)
	github.com/Dreamacro/clash/listener/http/proxy.go:32 +0x2c8
created by github.com/Dreamacro/clash/listener/http.NewWithAuthenticate.func1
	github.com/Dreamacro/clash/listener/http/server.go:61 +0x84
```

### Description

A panic exception occurred during 'docker run' startup with the current latest aarch64 version (>2021.07.03) of 'dreamacro/clash-premium' image.
The version '2021.07.03' and previous versions are normal.

Dreamacro:
@Kr328 cc

Dreamacro:
@Molay try latest image

Molay:
Thanks :)

PR #1555, opened 2021-08-20
Fix: HTTP proxy internal linkage signature

icpz:
This PR aims to fix https://github.com/Dreamacro/clash/issues/1552 .

The signature of the internal `net/http.readRequest` changed in go1.17, which breaks the call convention.


Dreamacro:
LGTM, thanks

PR #1505, opened 2021-07-23
Feature: add WebSocket early data

AkinoKaede:
It is compatible with V2Ray. 

Thanks to @ShinyGwyn

PR #1558, opened 2021-08-21
[Feature] support xray's ws-0rtt path scheme

wwqgtxx:
After #1505 is merged and #1473 is closed, add xray's ws-0rtt path scheme support and can work with v2ray-plugin without addition change.
based on : [xray's websocket document](https://xtls.github.io/config/transports/websocket/)

### DIFF
--- a/transport/vmess/websocket.go
+++ b/transport/vmess/websocket.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


Dreamacro:
I add a test for the xray docker image and ws 0rtt work great. So I checkout xray-core code and find that the server-side does not use this ed query.

https://github.com/XTLS/Xray-core/blob/d45298a10df50837aefbb7a1ec171fdd3fb767ab/transport/internet/websocket/hub.go#L48-L53

wwqgtxx:
Yes, this is a client-side behavior to match xray's configuration
> https://github.com/XTLS/Xray-core/pull/375
> 只需在客户端的 path 后加上 ?ed=2048 即可减少 1-RTT，2048 代表 early data 的长度上限，目前建议写 2048

The merged #1505 follow v2fly's config scheme which need xray server's user set like this:
```yaml
ws-path: '/'
ws-opts:
  max-early-data: 2048
  early-data-header-name: 'Sec-WebSocket-Protocol'
```
After this PR, we can simply use
```yaml
ws-path: '/?ed=2048'
```
It can work with any subscription conversion program (eg: https://github.com/tindy2013/subconverter ) without change.

Of course, this PR will NOT break up the v2fly's behavior, it still work fine which v2fly's server。


In addition, I read your test (https://github.com/Dreamacro/clash/commit/410772e81c1c2f212616d367bd0e25e2ec1508a8) and have a small suggestion:
Maybe split `TestClash_VmessWebsocket0RTT` to two function to test on both v2fly and xray server is better. 
And your `vmess-ws-0rtt.json`'s
```json
"wsSettings": {
  "maxEarlyData": 128,
  "earlyDataHeaderName": "Sec-WebSocket-Protocol"
}
```
only work with v2fly server but send this configure file to start a xray server. Certainly, a xray server will ignore those line. This test code may cause the reader which not clearly understand the different with v2fly and xray's implement to be confused

Dreamacro:
ok, thank you for your contribution

PR #1583, opened 2021-09-04
Feature: skip DIRECT proxies in relay

ExcitedCodes:
## What?
In relay outbound group, when there's a proxy unwrapped to `DIRECT` type, filter it out.
This PR also simplifies `ProxyAdapter.DialContext` procedure when only 1 proxy left after filtering.

## Why?
Imagine the following situation:

We got two routes to a proxy server **NodeA**:
 - **NodeA**, banned in some area but low cost
 - **NodeA - IPLC**, low-latency, not banned, high cost

(And we have a few more proxy servers similar to **NodeA**)

We also got a proxy service provider with some not banned servers, in this example:
 - Provider Node 1
 - Provider Node 2

Now we want to connect either via IPLC when we're surfing the web, or via that not banned proxy service provider when we want to download some large files:
 - Our Computer -> NodeA - IPLC
 - Our Computer -> Provider Node 1/2 -> NodeA

It's reasonable to come up with this config:
```yaml
proxy-groups:
- name: "Landing Node"
  type: select
  proxies:
  - NodeA
  - NodeA - IPLC
  - NodeB
  - NodeB - IPLC
  # and more ...
- name: "Optional Relay"
  type: select
  proxies:
  - DIRECT
  - Provider Node 1
  - Provider Node 2
  # and more ...
- name: "_relay"
  type: relay
  proxies:
  - "Optional Relay"
  - "Landing Node"
rules:
- "MATCH,_relay"
```

However, in current clash, using such config will lead to a connection error when we select `DIRECT` in `Optional Relay`:
```
error:  connect error: missing port in address
```

Which is caused by this relay handling way:
https://github.com/Dreamacro/clash/blob/7d20097465fc0e8d91b8a3f2434b471fcc144e3d/adapter/outboundgroup/relay.go#L28-L34

I've searched through issues/prs and found a staled feature request: https://github.com/Dreamacro/clash/issues/730
Therefore, I created this patch to solve error above, it simply skips DIRECT proxies, which meets (at least my) common sense.


adapter/outboundgroup/relay.go
```diff
@@ -21,10 +21,20 @@ type Relay struct {
...
+	var proxies []C.Proxy
+	for _, proxy := range r.proxies(metadata, true) {
+		if proxy.Type() != C.Direct {
+			proxies = append(proxies, proxy)
+		}
+	}
+
+	switch len(proxies) {
+	case 0:
+		return nil, errors.New("proxy does not exist or all proxies is DIRECT")
```
Dreamacro, line 33: all proxies **are** DIRECT

ExcitedCodes: Fixed, thanks

adapter/outboundgroup/relay.go
```diff
@@ -21,10 +21,20 @@ type Relay struct {
...
+			proxies = append(proxies, proxy)
+		}
+	}
+
+	switch len(proxies) {
+	case 0:
+		return nil, errors.New("proxy does not exist or all proxies are DIRECT")
+	case 1:
+		return proxies[0].DialContext(ctx, metadata)
 	}
```
icpz, line 36: How about 
```go
    switch len(proxies) {
    case 0:
        proxies = append(proxies, outbound.NewDirect())
        fallthrough
    case 1:
        return proxies[0].DialContext(ctx, metadata)
    }
```

This enables relay even when all proxies are DIRECT.

fengberd: That makes sense. I'd like to return directly to prevent extra memory allocation / move etc. Please check 
7ce552d

ExcitedCodes: That makes sense. I'd like to return directly to prevent extra memory allocation / move etc. Please check
7ce552d

Dreamacro:
Thanks for your contribution 👏

PR #1589, opened 2021-09-07
Fix: UDP leak if not match any rule

yaling888:
Fix UDP leak if not match any rule

### DIFF
--- a/tunnel/tunnel.go
+++ b/tunnel/tunnel.go
@@ -334,5 +334,8 @@ func match(metadata *C.Metadata) (C.Proxy, C.Rule, error) {
 		}
 	}
 
+	if metadata.NetWork == C.UDP {
+		return proxies["REJECT"], nil, nil
+	}
 	return proxies["DIRECT"], nil, nil
 }


fakeboboliu:
Well...I didn't get it. Why do you need to REJECT UDP w/o any match?

yaling888:
> Well...I didn't get it. Why do you need to REJECT UDP w/o any match?

Actually, it should be rejected instead of skipped in this place.
tunnel/tunnel.go, line 331
```
if metadata.NetWork == C.UDP && !adapter.SupportUDP() {
    log.Debugln("%s UDP is not supported", adapter.Name())
    return proxies["REJECT"], rule, nil
}
```

fakeboboliu:
https://github.com/Dreamacro/clash/blob/master/adapter/outbound/direct.go#L43

DIRECT supports UDP

yaling888:
The DIRECT adapter was used because the original matching UDP rule was skipped

yaling888:
When the server does not support UDP and the configuration file is set to support UDP in the proxy node, now you will find out a direct connection without a rule name on the dashboard, so UDP leaks.

fakeboboliu:
> When the server does not support UDP and the configuration file is set to support UDP in the proxy node, now you will find out a direct connection without a rule name on the dashboard, so UDP leaks.

Well…you're right, but is it worth to break connectivity by default in a tool that could be used as a gateway? 
Let's wait for @Dreamacro .

PR #1610, opened 2021-09-12
Fix: support leading zero padding in ip-cidr rule

fakeboboliu:
This PR intend to work with IP addresses like `001.001.001.002/32` in ip-cidr rules.

The zero padding is used in [RFC 870](https://datatracker.ietf.org/doc/html/rfc870) and [RFC 790](https://datatracker.ietf.org/doc/html/rfc790). It's better to provide support for robust.

As defined of dot-decimal notation, the leading zero should not be considered as a notation of octal, so removing them will not affect.

Since there's still usage of leading zeros in real-world (e.g. some firewalls, Nintendo switch), supporting them will reduce complexity to transform rules.

The leading zeroes can also provide an option to align ip-cidr rules for beautiful.

### DIFF
--- a/rule/ipcidr.go
+++ b/rule/ipcidr.go
@@ -2,6 +2,7 @@ package rules
 
 import (
 	"net"
+	"regexp"
 
 	C "github.com/Dreamacro/clash/constant"
 )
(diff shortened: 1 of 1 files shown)


fakeboboliu:
这问题确实太蠢了，还是关掉 PR 吧（

PR #1584, opened 2021-09-06
Feature: route based on HTTP/Socks authentication user

ExcitedCodes:
### STALE WARNING
This PR was inactive for a long time. If you're looking for binaries that supports this feature, use [clash-re](https://github.com/ExcitedCodes/clash/releases)

### What?
- Stores username used to authenticate (if enabled) into metadata while connecting via socks5/http listener
- Adds a new rule `AUTH-USER` to match username stored above

### Motivation
Clash is good at routing, especially with premium scripting feature & web dashboard to switch proxies instantly.
However, it lacks the ability to integrate with front proxy softwares (e.g. Proxifier, SwitchyOmega) currently, i.e. the front proxy software can't actively tell clash "this connection should be routed to a low-latency node / big-bandwidth node". All it can do is to choose connect directly or pass the traffic to clash.

1. Initially I'm creating this feature for per-app routing using Proxifier as the front proxy handler. Before you ask, the `PROCESS-NAME` rule wasn't powerful enough IMO. For example, Proxifier allows us to match every Adobe process using this pattern:
   ```
   [Applications]
   "C:\Program Files\Adobe\*"
   "C:\Program Files (x86)\Adobe\*"
   ```
   And it would be much better than writing a long, but may not match every request (and may match traffic from other programs) `IP-CIDR` & `DOMAIN-SUFFIX` list. Of course, we can create rule like `PROCESS-NAME-REGEX` to match these. But I believe it would be better to have clash running on a single device (e.g. the gateway) in order to maintain a single set of routing rule, and being able to do per-app routing on LAN devices at the same time.

2. I also found similar feature requests that can be satisfied with this PR:
    - #629 would like to route traffic based on protocol in order to "match all requests from SwitchyOmega"
    - #1242 would like to have multiple listeners and route specified port via a proxy group
    - #1568 would like to listen to another port in order to "route using chrome plugins", which is the "front proxy software"
    I've read through the listener mechanism of Clash, it would be a massive refactor to listen on multiple ports. Issues mentioned above can be solved elegantly if we're able to route by socks5/http user, which is well supported by most front proxy softwares.

### Implementation & Considerations
- Currently I'm storing authenticated user into metadata. The consideration is if this get merged, premium scripting can be boosted greatly.
- The `AuthUser` metadata won't be send to external controllers in API... Not sure if it's really needed.
- `tproxy` / `redir`: requests will have an empty `AuthUser` field, I believe If someone really want to distinguish these traffic, they'd better use scripting.
[...]

PR #1613, opened 2021-09-13
Fix: http/https proxy authentication

lxe524:
https://developer.mozilla.org/zh-CN/docs/Web/HTTP/Status/407
https://tools.ietf.org/html/rfc7235#section-3.2
According to above, "Proxy-Authenticate" header should not be remove when
the status is proxy authentication required.

This may fix Chrome SwitchyOmega directly use a https proxy include authentication.
Related to [1443](https://github.com/Dreamacro/clash/pull/1443) and [issuecomment-916173170](https://github.com/Dreamacro/clash/pull/1584#issuecomment-916173170)

### DIFF
--- a/listener/http/proxy.go
+++ b/listener/http/proxy.go
@@ -63,8 +63,8 @@ func HandleConn(c net.Conn, in chan<- C.ConnContext, cache *cache.Cache) {
 
 			request.RequestURI = ""
 
-			RemoveHopByHopHeaders(request.Header)
-			RemoveExtraHTTPHostPort(request)
+			removeHopByHopHeaders(request.Header)
+			removeExtraHTTPHostPort(request)
 
 			if request.URL.Scheme == "" || request.URL.Host == "" {
 				resp = responseWith(http.StatusBadRequest)
@@ -74,9 +74,9 @@ func HandleConn(c net.Conn, in chan<- C.ConnContext, cache *cache.Cache) {
 					resp = responseWith(http.StatusBadGateway)
 				}
 			}
-		}
 
-		RemoveHopByHopHeaders(resp.Header)
+			removeHopByHopHeaders(resp.Header)
+		}
 
 		if keepAlive {
 			resp.Header.Set("Proxy-Connection", "keep-alive")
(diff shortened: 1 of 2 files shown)


fakeboboliu:
The actual problem is `RemoveHopByHopHeaders` removed `Proxy-Authenticate`, but the solution should not be that straightforward, we still need `RemoveHopByHopHeaders` to work in all situations.

It will be better to solve this problem by adjusting the process order.

listener/http/proxy.go
```diff
@@ -76,7 +76,9 @@ func HandleConn(c net.Conn, in chan<- C.ConnContext, cache *cache.Cache) {
 			}
 		}
 
-		RemoveHopByHopHeaders(resp.Header)
+		if resp.StatusCode != http.StatusProxyAuthRequired {
+			RemoveHopByHopHeaders(resp.Header)
```
Dreamacro, line 80: seems that you can put `RemoveHopByHopHeaders` after line 76

BTW, `RemoveHopByHopHeaders` and `RemoveExtraHTTPHostPort` can be rename to `removeHopByHopHeaders` and `removeExtraHTTPHostPort`. Maybe you can modify it together?

lxe524: update a new solution, plz check. @fakeboboliu @Dreamacro

listener/http/proxy.go
```diff
@@ -36,7 +37,14 @@ func HandleConn(c net.Conn, in chan<- C.ConnContext, cache *cache.Cache) {
...
 
 		if !trusted {
-			resp = authenticate(request, cache)
+			resp, err = authenticate(request, cache)
+			if err != nil {
+				if resp.Write(conn) != nil {
+					break // close connection
+				}
+
+				return // authentication failure
```
Dreamacro, line 46: This will make the connection leak. Like I said before, put `removeHopByHopHeaders` after line 83 can solve the problem

https://github.com/Dreamacro/clash/blob/1ce43d1ab3193bb10d0031d1836bb5f3b7ce2b87/listener/http/proxy.go#L77-L84

lxe524: you are right, updated.

Dreamacro (approved):

Kr328 (approved):

Dreamacro:
@lxe524 thanks for your contribution 👏

PR #1617, opened 2021-09-13
Fix: DNS client ExchangeWithConn should not use outer err

Kr328:
尝试修正一个可能在极端条件下 msg 和 err 都为 nil 的问题


Kr328:
https://github.com/Dreamacro/clash/commit/55600c49c9299744c2c3d67dc72ac0d798bc3437

PR #1624, opened 2021-09-15
Fix: socks4 request continues after authentication failed

ExcitedCodes:
### What?
 - Fixes a socks4 authentication bug (the request continues even if auth failed and RequestIdentdMismatched was sent)

(separated from https://github.com/Dreamacro/clash/pull/1584)


transport/socks4/socks4.go
```diff
@@ -99,7 +100,11 @@ func ServerHandshake(rw io.ReadWriter, authenticator auth.Authenticator) (addr s
 	copy(reply[4:8], dstIP)
 	copy(reply[2:4], dstPort)
 
-	_, err = rw.Write(reply[:])
+	if err == nil {
```
Dreamacro, line 103: It doesn't look so clear here.

```go
_, wErr := rw.Write(reply[:])
if err != nil {
    return err
}
return wErr
```

Do you think that would be better?

ExcitedCodes: That snippet won't compile... It's named return vars. See f4a415c4f659fc4070276b5e139e7a6e130e887a.

Dreamacro:
Thanks! 👏

PR #1630, opened 2021-09-17
Fix: return correct DNS cookie

w1ndy:
Clash does not correctly respond to [DNS cookie](https://kb.isc.org/docs/aa-01387) challenges because of caches and parallel requests. Try testing with dig (which cookies are on by default in the recent versions):

```bash
# Do this for multiple times because sometimes the first response has a valid cookie
$ dig @localhost -p 1099 +cookie www.baidu.com
...
;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; MBZ: 0x03fc, udp: 4096
; COOKIE: 10c376211e65d91640666c0c61447389c864fd87caf2013e (bad)
;; QUESTION SECTION:
...
```

A minimal configuration to reproduce this bug:
```yaml
dns:
  enable: true
  listen: 0.0.0.0:1099
  nameserver: ['119.29.29.29']
```

This commit adds correct cookies to the DNS responses.

### DIFF
--- a/dns/resolver.go
+++ b/dns/resolver.go
@@ -90,6 +90,42 @@ func (r *Resolver) Exchange(m *D.Msg) (msg *D.Msg, err error) {
 	return r.ExchangeContext(context.Background(), m)
 }
 
+func (r *Resolver) extractDnsCookie(m *D.Msg) (string, error) {
+	if edns0 := m.IsEdns0(); edns0 != nil {
+		for _, opt := range edns0.Option {
+			if opt.Option() == D.EDNS0COOKIE {
+				return opt.(*D.EDNS0_COOKIE).Cookie, nil
+			}
+		}
+	}
+	return "", errors.New("no dns cookie")
+}
+
+func (r *Resolver) syncDnsCookie(request *D.Msg, response *D.Msg) {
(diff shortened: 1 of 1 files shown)


dns/resolver.go
```diff
@@ -90,6 +90,42 @@ func (r *Resolver) Exchange(m *D.Msg) (msg *D.Msg, err error) {
...
+		}
+	}
+	return "", errors.New("no dns cookie")
+}
+
+func (r *Resolver) syncDnsCookie(request *D.Msg, response *D.Msg) {
+	if cookie, err := r.extractDnsCookie(request); err == nil {
+		c := new(D.EDNS0_COOKIE)
+		c.Code = D.EDNS0COOKIE
+		c.Cookie = cookie + "deadbeefdeadbeef"
```
fakeboboliu, line 108: > `"deadbeefdeadbeef"`

Are you debugging?

w1ndy: This is the [server cookie](https://pkg.go.dev/github.com/miekg/dns?utm_source=godoc#EDNS0_COOKIE), which should be at least 8 bytes and can be arbitrary. I found using a magic string very useful because you can tell whether the dns responses are coming from clash or not.

PR #1633, opened 2021-09-20
Feature: Add country and process name into metadata

wotmshuaisi:
image


Dreamacro:
https://github.com/Dreamacro/clash/pull/1486#issuecomment-877621099

PR #1650, opened 2021-09-28
Improve: avoid bufconn twice

fakeboboliu:
This PR was a part of #1603.

This PR intend to avoid recreating on an existing but shadowed BufferedConn.

For instance:

https://github.com/Dreamacro/clash/blob/dev/listener/mixed/mixed.go#L66-L81

creates a bufconn, and passed into `http.HandleConn`, which makes a bufconn on this bufconn again.

To prevent misuses in the future, I unshadowed bufconn instead of adding `NewBufferedConn` to other callers of `http.HandleConn` and removing `NewBufferedConn` inside.


Dreamacro:
Thanks!

PR #1719, opened 2021-11-08
Fix: dial-options config parse

fakeboboliu:
This PR was a part of #1705.

This PR does:
 - Remove the requirement of interface-name in the config, which should be optional
 - Make interface-name affect group
 - Fix config parsing when anonymous embed is used


Dreamacro:
e622d8d Seems to work better

PR #1730, opened 2021-11-11
updated healthcheck, added param concurrency.

norci:
the hard coded value is 10, which is too small, for a large proxy
list.

I have tested it. it can finish health check in a minute, for a list contains 20,000 proxys, with concurrency value 1000.


norci:
It seems https://github.com/Dreamacro/clash/pull/1535 is a better solution

PR #1821, opened 2021-12-12
Proxygroup always consider health-check options in mixed configurations

noel2004:
Consider following configurations for a proxy group which mixed both proxy-provider and proxies:

```yaml
- name: "auto"
  type: url-test
  use:
    - provider
  proxies:
    - ss1
    - ss2
    - vmess1
    # tolerance: 150
  url: 'http://www.gstatic.com/generate_204'
  interval: 300
```

In issue #794, it is said that the proxies would be wrapped as a provider and the health-check options (url / interval) is used. But current in line 62 of adapter/outboundgroup/parser.go the options is dropped when proxy-provider is also involved. 

I am not very understand why this dropping is necessary. And if this would be considered to be corrected then this PR may help.


Dreamacro:
It looks like your changes are reasonable, thanks

PR #1871, opened 2022-01-03
Add rule support for domain prefix

cbakgly:
The way "domain suffix" does likewise (thanks to ^c - ^v skill:-).
Local test passed. 

Ps: prefix match would be very useful to block many ads/tracking request.


ruisiji:
Is this really needed?

Clash already supported `DomainKeyword`, this can cover your usage(block ads/tracking domains).

cbakgly:
> Is this really needed?
> 
> Clash already supported `DomainKeyword`, this can cover your usage(block ads/tracking domains).

Hi,
Sorry for late reply. Just happened to check mailbox and saw this update.
Well, in my matching list, there are quite amount of advertising URLs with some pattern like starting with ad[0-9]*.
And maybe more of them out there but not in the list. 
Personally I think prefix is an easy way to catch all of such URLs, and also make the list clean and tidy.

PR #1893, opened 2022-01-09
fix tls handshake unexpected timeout.

thank243:
This a temporary solution. Waiting official fixed it.


transport/gun/gun.go
```diff
@@ -173,6 +174,12 @@ func NewHTTP2Client(dialFn DialFn, tlsConfig *tls.Config) *http2.Transport {
 		}
 
 		cn := tls.Client(pconn, cfg)
+
+		// fix tls handshake not timeout
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		_ = cn.HandshakeContext(ctx)
+
 		if err := cn.Handshake(); err != nil {
```
icpz, line 183: `HandshakeContext` will perform the actual handshake, so we should check it's error and remove the unnecessary `Handshake`:

```go
if err := cn.HandshakeContext(ctx); err != nil {
...
}
```

transport/gun/gun.go
```diff
@@ -173,6 +174,12 @@ func NewHTTP2Client(dialFn DialFn, tlsConfig *tls.Config) *http2.Transport {
 		}
 
 		cn := tls.Client(pconn, cfg)
+
+		// fix tls handshake not timeout
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
icpz, line 179: We can add a constant `DefaultTLSTimeout = 10 * time.Second` in `constant/adapters.go`, and reference it here:
`ctx, cancel := context.WithTimeout(context.Background(), constant.DefaultTLSTimeout)`

icpz (changes requested):

thank243:
add TLS timeout const in constant.

transport/gun/gun.go
```diff
@@ -5,19 +5,20 @@ package gun
 
 import (
 	"bufio"
+	"context"
 	"crypto/tls"
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"github.com/Dreamacro/clash/common/pool"
+	C "github.com/Dreamacro/clash/constant"
```
icpz, line 14: Sorry but please don't change the order and empty lines of the imports.

icpz (changes requested):

icpz (approved):
LGTM

Dreamacro:
Thanks!

PR #2028, opened 2022-03-16
fix routing-mark option doesn't work on shadowsocks

risetechlab:
Hi,
I found that the routing-mark option under shadowsocks doesn't make any effect. This pull request fixed it.


Dreamacro:
It seems that many others outbound proxies are missing this. Can you fix it together?

risetechlab:
> It seems that many others outbound proxies are missing this. Can you fix it together?

Hi,
Sorry for keeping you waiting. The latest commit fixed all the other protocols.

Dreamacro:
nice work, thanks for your contribution.

PR #2041, opened 2022-03-22
Fix: convert size to unit32 in function getoridst to fix EINVAL on mips64

maomaozgw:
Fix getorigdst cannot get origin destination address in MIPS64 (big-endian) in linux. 

Because unsafe.Sizeof returns uintptr (8 byte) and kernel use int (4 byte) to receive data size, in big-endian the value will be truncated to zero, kernel check data length failed with EINVAL.


Dreamacro:
@Kr328 /cc

Dreamacro:
@maomaozgw thanks

PR #2042, opened 2022-03-23
When processing data through `json.Unmarshal`, `port` thinks it is `float64`, which causes the parsing node to report an error

Luoxin:
When processing data through `json.Unmarshal`, `port` thinks it is `float64`, which causes the parsing node to report an error

### DIFF
--- a/common/structure/structure.go
+++ b/common/structure/structure.go
@@ -91,6 +91,8 @@ func (d *Decoder) decodeInt(name string, data any, val reflect.Value) (err error
 	switch {
 	case kind == reflect.Int:
 		val.SetInt(dataVal.Int())
+	case kind == reflect.Float64:
+		val.SetInt(int64(dataVal.Float()))
 	case kind == reflect.String && d.option.WeaklyTypedInput:
 		var i int64
 		i, err = strconv.ParseInt(dataVal.String(), 0, val.Type().Bits())
--- a/common/structure/structure.go
+++ b/common/structure/structure.go
@@ -91,7 +91,7 @@ func (d *Decoder) decodeInt(name string, data any, val reflect.Value) (err error
 	switch {
 	case kind == reflect.Int:
 		val.SetInt(dataVal.Int())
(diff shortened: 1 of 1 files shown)


common/structure/structure.go
```diff
@@ -91,6 +91,8 @@ func (d *Decoder) decodeInt(name string, data any, val reflect.Value) (err error
 	switch {
 	case kind == reflect.Int:
 		val.SetInt(dataVal.Int())
+	case kind == reflect.Float64:
+		val.SetInt(int64(dataVal.Float()))
 	case kind == reflect.String && d.option.WeaklyTypedInput:
```
Dreamacro, line 96: Float64 and String belong to `WeaklyTypedInput`

Luoxin: When I use 
`{"alterId":2,"cipher":"auto","name":"proxy","network":"ws","port":15018,"server":"127.0.0.1" ,"tls":false,"type":"vmess","uuid":"047184b7-6da2-3d3f-ac27-6a1a8701daf8","ws-opts":{"headers":{"Host":"127.0. 0.1"},"path":"/"}}` 
after doing `json.Unmarshal`, as `adapter.ParseProxy` input, it returns me an error 
`'port' expected type 'int', got unconvertible type 'float64'`

image

Dreamacro: The consequences of int -> float caused by golang json deserialization should not be borne by `structure`.  `Float64` 
 should have the same WeaklyTypedInput determination as `String`

Luoxin: Developers will inevitably experience such an error before knowing that golang json deserialization will cause this problem. At the same time, there is more than one `port` field of `int` type, which means that developers need to deal with some third-party packages Resolved compatibility issues during their ongoing testing.

Luoxin: Maybe I misunderstood the meaning, you mean like this?
image

Dreamacro:
Thanks 👏

PR #2054, opened 2022-04-03
Feature: Increate NOFILE limit if we can, because why not?

fakeboboliu:
Like ss-rust done in their project. Just automatically increase soft limit to hard limit.

It costs nothing, and will avoid some potential issues.


fakeboboliu:
Wellllllllllllllll

https://go-review.googlesource.com/c/go/+/393354

PR #2056, opened 2022-04-04
Add full support for SSR

Anankke:
ref https://github.com/Dreamacro/go-shadowsocks2/pull/4


Dreamacro:
Why not convert `none` to `dummy`?

Anankke:
There is no such thing called `dummy` in SSR protocol; converting this breaks SSR protocol compatibility.

Dreamacro:
```go
if cipher == "none" {
   cipher = "dummy" 
}
```

Anankke:
Both work, which do you prefer?

Dreamacro:
I prefer not to change go-shadowsocks2

Anankke:
Are we good now?

Dreamacro:
Thanks for contributing, and for your patience

Anankke:
NP ala its fixed thank you

PR #2066, opened 2022-04-10
Fix: custom hosts should be added to mapping table during the DNS query

yaling888:
Custom hosts should be added to mapping table during the DNS query, that can be restored it when pre handle metadata.


Dreamacro:
Please describe what problems have been met.

yaling888:
config:
```yaml
hosts:
  'example.com': 'x.x.x.x'

dns:
  enable: true
  use-hosts: true
  listen: 0.0.0.0:53
  enhanced-mode: redir-host # or fake-ip

rules:
  - DOMAIN,example.com,proxy
```
problem is that host name 'example.com' cannot match rule DOMAIN.

Dreamacro:
This causes a problem similar to redir-host, where a single ip (e.g. 127.0.0.1) will mapping wrong host if it corresponds to multiple host domains

yaling888:
> This causes a problem similar to redir-host, where a single ip (e.g. 127.0.0.1) will mapping wrong host if it corresponds to multiple host domains

Reducing TTL may alleviate in your case

PR #2104, opened 2022-04-28
Chore(build): add linux-loong64

abner-chenc:
Signed-off-by: Guoqi Chen <chenguoqi@loongson.cn>


Dreamacro:
It looks like go 1.18.1 doesn't support `GOARCH=loong64` yet

PR #2126, opened 2022-05-12
Feature: add specific interface for nameserver

Bowser1704:
Changing nameserver schema.

`{protocal}://{ip}:{port}@{interface}`

So, clash can connect private `dns` server through other vpn.

e.g. `10.0.0.1@vpn`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -468,6 +468,14 @@ func parseNameServer(servers []string) ([]dns.NameServer, error) {
 	nameservers := []dns.NameServer{}
 
 	for idx, server := range servers {
+		// parse with specific interface
+		// .e.g 10.0.0.1@vpn
+		interfaceName := ""
+		if strings.Contains(server, "@") {
+			s := strings.Split(server, "@")
+			server, interfaceName = s[0], s[1]
+		}
+
 		// parse without scheme .e.g 8.8.8.8:53
 		if !strings.Contains(server, "://") {
 			server = "udp://" + server
@@ -506,8 +514,9 @@ func parseNameServer(servers []string) ([]dns.NameServer, error) {
 		nameservers = append(
 			nameservers,
 			dns.NameServer{
-				Net:  dnsNetType,
-				Addr: addr,
+				Net:       dnsNetType,
+				Addr:      addr,
+				Interface: interfaceName,
 			},
 		)
 	}


Dreamacro:
Using the standard url format should be better.

example:
```
tls://en0@8.8.8.8
tls://8.8.8.8#en0
tls://8.8.8.8?interface=en0
```

config/config.go
```diff
@@ -468,6 +468,14 @@ func parseNameServer(servers []string) ([]dns.NameServer, error) {
 	nameservers := []dns.NameServer{}
 
 	for idx, server := range servers {
+		// parse with specific interface
```
Dreamacro, line 471: It should not be parsed manually, but should be retrieved directly from the parsed url object

Dreamacro:
@Kr328 @icpz What are your thoughts on where to put the interface? querystring looks more expandable, but also more verbose. The other two are not conducive to expansion but simple.

icpz:
Will the querystring format messes up with doh servers? e.g.
```
https://doh.pub/dns-query?interface=en0
```
It seems kind of weird in such case.

icpz:
The pound sign form can also be expandable, say `tls://8.8.8.8#iface=en0#foo=bar#para=arg`, seems ugly though

config/config.go
Bowser1704: excuse me， I don't get your point, do you mean to modify the comment?

Dreamacro: This url is parsed at the bottom, using u.Fragment instead.

Dreamacro:
@icpz
For DoH, will create a new url, so there is no such problem.

> The pound sign form can also be expandable, say tls://8.8.8.8#iface=en0#foo=bar#para=arg, seems ugly though

url only has one fragment. The question is whether the `interface` should be special?

icpz:
I'd like a `interface` for explicitness and expand extensibility

config/config.go
Bowser1704: get

Bowser1704:
> I'd like a `interface` for explicitness and expand extensibility

I think `interface` should be special, using query string is a little weird.

Dreamacro:
@Bowser1704 DoH current implementation does not contain `interface`, should it be added?

Bowser1704:
> @Bowser1704 DoH current implementation does not contain `interface`, should it be added?

I haven't used DoH. Do you think it should be added?

Dreamacro:
yes, DoH is part of dns client, and the feature is `add specific interface for nameserver`

Bowser1704:
@Dreamacro maybe we should add some nameserver with specific interface in `dns_test.go`, however, we should add some appropriate interface name for GitHub action.

dns/doh.go
```diff
@@ -79,12 +81,18 @@ func (dc *dohClient) doRequest(req *http.Request) (msg *D.Msg, err error) {
 	return msg, err
 }
 
-func newDoHClient(url string, r *Resolver) *dohClient {
+func newDoHClient(dnsUrl string, r *Resolver) *dohClient {
```
Dreamacro, line 84: 网卡参数直接传进函数就行

Dreamacro:
Thank!

PR #2199, opened 2022-07-01
Change: remove AddrType on Metadata

Dreamacro:
Remove unnecessary `AddrType` field.

Need to evaluate this change together.


icpz (approved):
LGTM

PR #2206, opened 2022-07-05
Chore: fix process rule type

kamingchan:


rule/process.go
```diff
@@ -14,7 +14,11 @@ type Process struct {
 }
 
 func (ps *Process) RuleType() C.RuleType {
-	return C.Process
+	if ps.nameOnly {
+		return C.Process
+	} else {
```
Dreamacro, line 19: I prefer not `else`

kamingchan: Fixed.

Dreamacro:
thanks a lot 👍

PR #2220, opened 2022-07-12
Fix: handle parse socks5 udp address properly

kamingchan:
Seems to be a common mistake in Go.

```go
func a() *net.UDPAddr {
	return nil
}

func b() net.Addr {
	return a()
}

func Test(t *testing.T) {
	b := b()
	_, ok := b.(*net.UDPAddr)
	assert.True(t, b != nil)
	assert.True(t, ok)
}
```

[`socsk5.Addr.UDPAddr()`](https://github.com/Dreamacro/clash/blob/f0952b55d01c6d57b49758d63eef33b3538b54cc/transport/socks5/socks5.go#L70) should check nil properly.  After converted to `net.Addr` in `ReadFrom(p []byte) (n int, addr Addr, err error)`, nil value of `*net.UDPAddr` is always not equal to nil but [calling its method may cause panic](https://github.com/Dreamacro/clash/blob/f0952b55d01c6d57b49758d63eef33b3538b54cc/tunnel/connection.go#L46) (its type is not nil but value is nil).


Dreamacro:
on snell.go, it does not panic (there is a relevant check earlier). But adding a check is also necessary

PR #2234, opened 2022-07-23
add HTTPS support for external controller

khronosmoe:
>The site http://yacd.haishan.me/ is served with HTTP not HTTPS is because many browsers block requests to HTTP resources from a HTTPS website. If you think it's not safe, you could just download the [zip of the gh-pages](https://github.com/haishanh/yacd/archive/gh-pages.zip), unzip and serve those static files with a web server(like Nginx).

In order to solve this, I make this commit to add HTTPS support for external controller. Config is as follows:
```yaml
external-controller: 127.0.0.1:9090
external-ui: yacd
secret: ""
external-tls: true
tls-cert-file: path_to_cert
tls-key-file: path_to_key
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -53,6 +53,9 @@ type Controller struct {
 	ExternalController string `json:"-"`
 	ExternalUI         string `json:"-"`
 	Secret             string `json:"-"`
+	ExternalTLS        bool   `json:"-"`
+	TLSCertPath        string `json:"-"`
+	TLSKeyPath         string `json:"-"`
 }
 
 // DNS config
@@ -137,6 +140,9 @@ type RawConfig struct {
 	ExternalController string       `yaml:"external-controller"`
 	ExternalUI         string       `yaml:"external-ui"`
 	Secret             string       `yaml:"secret"`
+	ExternalTLS        bool         `yaml:"external-tls"`
+	TLSCertPath        string       `yaml:"tls-cert-file"`
+	TLSKeyPath         string       `yaml:"tls-key-file"`
 	Interface          string       `yaml:"interface-name"`
 	RoutingMark        int          `yaml:"routing-mark"`
 
@@ -252,6 +258,7 @@ func parseGeneral(cfg *RawConfig) (*General, error) {
 		}
 	}
 
+
 	return &General{
 		Inbound: Inbound{
 			Port:        cfg.Port,
@@ -266,6 +273,9 @@ func parseGeneral(cfg *RawConfig) (*General, error) {
 			ExternalController: cfg.ExternalController,
 			ExternalUI:         cfg.ExternalUI,
 			Secret:             cfg.Secret,
+			ExternalTLS:        cfg.ExternalTLS,
+			TLSCertPath:        cfg.TLSCertPath,
+			TLSKeyPath:         cfg.TLSKeyPath,
 		},
 		Mode:        cfg.Mode,
 		LogLevel:    cfg.LogLevel,
--- a/hub/hub.go
+++ b/hub/hub.go
@@ -26,6 +26,24 @@ func WithSecret(secret string) Option {
 	}
 }
 
+func WithExternalTLS(externalTLS bool) Option {
+	return func(cfg *config.Config) {
+		cfg.General.ExternalTLS = externalTLS
+	}
+}
+
+func WithTLSCertPath(tLSCertPath string) Option {
+	return func(cfg *config.Config) {
+		cfg.General.TLSCertPath = tLSCertPath
+	}
+}
+
+func WithTLSKeyPath(tLSKeyPath string) Option {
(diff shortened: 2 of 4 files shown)


Dreamacro:
I think TLS offloading should be done by nginx or caddy, not on clash

khronosmoe:
Check https://blog.thris.me/archives/cldash-https.html , it would be better to provide a direct https function for RESTful API and dashboard than to use nginx. Meanwhile, Surge now also [supports](https://manual.nssurge.com/others/misc-options.html#http-api-tls).

khronosmoe:
To make my point clear, it is overkill and complicated to use Nginx to reverse proxy both the RESTful API and static web dashboard, considering most use cases are very simple just a web hosting, which the golang build-in ServeTLS() function is enough. And this is an option, users can choose whether to use Nginx or built-in function.

btwiuse:
> > The site http://yacd.haishan.me/ is served with HTTP not HTTPS is because many browsers block requests to HTTP resources from a HTTPS website. If you think it's not safe, you could just download the [zip of the gh-pages](https://github.com/haishanh/yacd/archive/gh-pages.zip), unzip and serve those static files with a web server(like Nginx).
> 
> In order to solve this, I make this commit to add HTTPS support for external controller. Config is as follows:
> 
> ```yaml
> external-controller: 127.0.0.1:9090
> external-ui: yacd
> secret: ""
> external-tls: true
> tls-cert-file: path_to_cert
> tls-key-file: path_to_key
> ```

My solution to the problem is to allow setting `external-ui` to a remote url, like
```
external-ui: https://clashboard.vercel.app
```

Instead of embedding static assets into the binary, it serves the web ui by reverse-proxying a remote url.

Pros:
- you don't have to manually download ui anymore
- you don't need to separately handle the ui directory when configuring nginx, caddy etc.
- no more `https://webui -> http://clash` problem, the webui is directly accessible from `http://clash`

Cons:
- needs external network access

You can try my fork at https://github.com/btwiuse/dash

I can make a pull request if this idea sounds good to you. @Dreamacro

khronosmoe:
> > > The site http://yacd.haishan.me/ is served with HTTP not HTTPS is because many browsers block requests to HTTP resources from a HTTPS website. If you think it's not safe, you could just download the [zip of the gh-pages](https://github.com/haishanh/yacd/archive/gh-pages.zip), unzip and serve those static files with a web server(like Nginx).
> > 
> > 
> > In order to solve this, I make this commit to add HTTPS support for external controller. Config is as follows:
> > ```yaml
> > external-controller: 127.0.0.1:9090
> > external-ui: yacd
> > secret: ""
> > external-tls: true
> > tls-cert-file: path_to_cert
> > tls-key-file: path_to_key
> > ```
> 
> My solution to the problem is to allow setting `external-ui` to a remote url, like
> 
> ```
> external-ui: https://clashboard.vercel.app
> ```
> 
> Instead of embedding static assets into the binary, it serves the web ui by reverse-proxying a remote url.
> 
> Pros:
> 
>     * you don't have to manually download ui anymore
> 
>     * you don't need to separately handle the ui directory when configuring nginx, caddy etc.
> 
>     * no more `https://webui -> http://clash` problem, the webui is directly accessible from `http://clash`
> 
> 
> Cons:
> 
>     * needs external network access
> 
> 
> You can try my fork at https://github.com/btwiuse/dash
> 
> I can make a pull request if this idea sounds good to you. @Dreamacro

What is the connection between your Clash and the remote server? Is it unencrypted? By the way, it is not " wise " to make the controller public and I strongly believe that the ServeTLS() function is a good and simple way.

btwiuse:
> What is the connection between your Clash and the remote server? Is it unencrypted? By the way, it is not " wise " to make the controller public and I strongly believe that the ServeTLS() function is a good and simple way.

My fork clash (called dash onward) allows starting clash like this
```
./clash -ext-ctl 127.0.0.1:9090 -ext-ui https://clashboard.vercel.app
```

When you visit http://127.0.0.1:9090, dash will start a reverse proxy with https://clashboard.vercel.app being the upstream url

any requests not matching
- /version
- /logs
- /traffic
- /configs
- /proxies
- /rules
- /connections
- /providers/proxies

will fallback to the reverse proxy handler

This is effectively the same as serving the webui on the root path http://127.0.0.1:9090/, and does not require making the controller public.

khronosmoe:
> This is effectively the same as serving the webui on the root path http://127.0.0.1:9090/, and does not require making the controller public.
> https://clashboard.vercel.app/ does not have an associated clash backend and won't leak your private info to the public.

Okay, I get it, but I'm not going to use it in my production. Considering that there are already static files like rulesets and mmdb under the .config folder, why not put the dashboard there too? I appreciate laggardkernel's [brew formula](https://github.com/laggardkernel/homebrew-tap/blob/master/Formula/clash-premium.rb)

PR #2246, opened 2022-07-30
refactor: update Dockerfile

adriang-90:
* Instead of separating the Go module download, we could remove this and just use a cache mount for ```/go/pkg/mod```
* ```COPY``` will create an extra layer in the container image which slows things down and uses extra disk space. This can be avoided by using ```RUN --mount``` and bind mounting from the build context, from a stage, or an image.
* add a cross compiling targets to the Dockerfile: ```TARGETOS``` & ```TARGETARCH```
* add ```BUILDPLATFORM``` -> matches current machines CPU architecture (eg. linux/amd64)
* go build reduced by 1/3 and the image size has been reduced by 1/4 (22MB->16MB)

Useful links:
https://www.docker.com/blog/containerize-your-go-developer-environment-part-1/
https://www.reddit.com/r/golang/comments/q7zppz/docker_cache_for_dependencies/
https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,19 +2,17 @@ FROM --platform=${BUILDPLATFORM} golang:alpine as builder
 
 RUN apk add --no-cache make git ca-certificates tzdata && \
     wget -O /Country.mmdb https://github.com/Dreamacro/maxmind-geoip/releases/latest/download/Country.mmdb
-WORKDIR /clash-src
+WORKDIR /workdir
 COPY --from=tonistiigi/xx:golang / /
-ARG TARGETOS TARGETARCH
+ARG TARGETOS TARGETARCH TARGETVARIANT
 
 RUN --mount=target=. \
     --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=cache,target=/go/pkg/mod \
-    GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
-    CGO_ENABLED=0 go build -trimpath -ldflags '-X "github.com/Dreamacro/clash/constant.Version=$(VERSION)" \
-		-X "github.com/Dreamacro/clash/constant.BuildTime=$(BUILDTIME)" \
-		-w -s -buildid=' -o /clash
+    make BINDIR= ${TARGETOS}-${TARGETARCH}${TARGETVARIANT} && \
+    mv /clash* /clash
 
-FROM scratch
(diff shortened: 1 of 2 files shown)


Dreamacro:
There are some changes that look cool, but for some of them I would like to say my opinion.

* I'm using github actions to build images and it looks like the current changes don't support armv6 and armv7
https://github.com/Dreamacro/clash/blob/8c7c8f43741923d6a7650dc723b4b51b0f540777/.github/workflows/docker.yml#L49
* Whether these changes (like caching and these default parameters) are not used in other OCI-compatible implementations, such as https://github.com/containers/buildah.
* Is it possible to build directly with makefile. Since docker builds from go, their description of the platform is similar. I made a simple change here to make it support reusing the makefile.
```dockerfile
FROM --platform=${BUILDPLATFORM} golang:alpine as builder

RUN apk add --no-cache make git ca-certificates tzdata && \
    wget -O /Country.mmdb https://github.com/Dreamacro/maxmind-geoip/releases/latest/download/Country.mmdb
WORKDIR /workdir
COPY --from=tonistiigi/xx:golang / /
ARG TARGETOS TARGETARCH TARGETVARIANT

RUN --mount=target=. \
    --mount=type=cache,target=/root/.cache/go-build \
    --mount=type=cache,target=/go/pkg/mod \
    make BINDIR= ${TARGETOS}-${TARGETARCH}${TARGETVARIANT} && \
    mv /clash* /clash

FROM alpine:latest
LABEL org.opencontainers.image.source="https://github.com/Dreamacro/clash"

COPY --from=builder /usr/share/zoneinfo /usr/share/zoneinfo
COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
COPY --from=builder /Country.mmdb /root/.config/clash/
COPY --from=builder /clash /
ENTRYPOINT ["/clash"]
```
* Since this is a network-related application, there will be times when you run into problems and need to go into the container to troubleshoot the problem. Alpine linux does a good job of providing a simple shell environment and package management tools `apk` to install packages such as curl, netcat, etc.
So although using scratch brings some size improvements, I would prefer to use alpine.

Dreamacro:
No response for a long time, so I will push my changes to this PR in a few days and merge it

PR #2247, opened 2022-07-30
Pass run golangci-lint-1.47.2

yaling888:
Pass run golangci-lint with version number v1.47.2


Dreamacro:
I want to keep the existing sorting, but the new version of gci "breaks" it, and I want to wait until gci supports custom sorting, or switch to another sorting tool.

Dreamacro:
https://github.com/daixiang0/gci/releases/tag/v0.6.0 seems gci support custom order

Dreamacro:
fixed by 6648793, latest `golangci-lint` support gci custom import order

PR #2248, opened 2022-08-01
Fix: auto health check

kamingchan:


Dreamacro:
The original logic did not seem to be a problem. When in `lazy` state, the health check is performed only after touch in this time interval.

kamingchan:
OK. I misunderstand the defination of `lazy`. :)

PR #2260, opened 2022-08-11
Fix: Satisfy DNS Case Insensitivity

fakeboboliu:
This PR is to implement "a lookup string octet with a value in the inclusive range from 0x41 to 0x5A, the uppercase ASCII letters, MUST match the identical value and also match the corresponding value in the inclusive range from 0x61 to 0x7A, the lowercase ASCII letters." in RFC4343.

Which could mitigate massive DNS records produced by software which changes case to avoid DNS cache and to track users.

图片


component/fakeip/pool.go
```diff
@@ -38,6 +39,11 @@ func (p *Pool) Lookup(host string) net.IP {
 	defer p.mux.Unlock()
 	if ip, exist := p.store.GetByHost(host); exist {
```
Dreamacro, line 40: why not `strings.ToLower` before GetByHost. And It would be good to add a comment on why this was added

fakeboboliu: > why not `strings.ToLower` before GetByHost. And It would be good to add a comment on why this was added

Call `strings.ToLower` before GetByHost will cause a slightly performance lost since most query is already lowercase, but it's not a big deal, I changed it for code style now.

Dreamacro:
Thanks, I found some information about this. It looks like it should be transformed to lower case on fakeip.

https://datatracker.ietf.org/doc/html/draft-vixie-dnsext-dns0x20-00
https://www.ietf.org/proceedings/72/slides/dnsext-2.pdf

PR #2325, opened 2022-09-26
fix uncorrect README wiki link

Roiocam:
README wiki is not quite correct, It does not a bug report issue and I could pull a request to fix this.


Dreamacro:
Thanks 🙌

PR #2331, opened 2022-09-28
Upgrade Go version to 1.19 that missing upgrade before

A7103:
Hello, developer, I found that you upgraded the version of Go before ([commit](https://github.com/Dreamacro/clash/commit/50105f0559f514b308e6997163cb6da99ae6b9fe)) , but you seem to have forgotten something, I add the rest it in here


Dreamacro:
When pushing this commit, the minimum version supports go 1.18.

Now that some go 1.19 APIs have been added that cause builds to fail on older versions, I think it's reasonable to limit it to 1.19 due to the obvious memory allocation improvements observed with clash on go 1.19.

PR #2378, opened 2022-10-31
Fix: a shared fastSingle.Do() may cause providers untouched

wwqgtxx:
In `URLTest`, `fastSingle` will cache the result in 10s. If there are many request call `fast(false)`, maybe the `u.proxies()` inside will never be called by `true`.


Dreamacro:
nice catch

PR #2398, opened 2022-11-07
Add back go1.18 support

nekohasekai:
Some people still need this stable version that is still supported by Go.


Dreamacro:
In my actual use, go 1.19 is a big improvement in memory control, and I'm not quite sure what the reason is for not setting the minimum version to 1.19

PR #2437, opened 2022-11-26
add build for android

URenko:
Now we can run clash directly in Android like v2ray-core

See https://guide.v2fly.org/app/android_without_root_use_v2ray.html for usage. I have checked that it works.


Dreamacro:
As far as I know, it is possible to execute `GOOS=linux` binary directly from the Android terminal. Are there any special benefits to using adb shell execution?

PR #2440, opened 2022-11-28
Chore: allow tunnels don't set special proxy

wwqgtxx:


Dreamacro:
It looks like it's throwing the tunnel request back to the rule match, which doesn't seem like good behavior. Are there any usage scenarios?

wwqgtxx:
Maybe some users just want to forward the port of a remote server to the local, and don't want to specify a specific proxy or proxy group.

Dreamacro:
But most of the rules are for the destination address (Host), and there seems to be unnecessary in throwing the request back into the rules to match

wwqgtxx:
Many users are using rule-providers or distribution rules provided by third parties such as airports/subscription conversions. They may want to simply configure tunnels to forward ports and automatically apply the previous rules.

PR #2463, opened 2022-12-13
fix: safeConnClose not working

wwqgtxx:
follow this link: https://stackoverflow.com/a/42703862
>  If you pass any parameters to the defer func (like the 2nd defer function above), those parameters are evaluated when the defer function is deferred, not when they are executed.

### DIFF
--- a/adapter/outbound/trojan.go
+++ b/adapter/outbound/trojan.go
@@ -131,17 +131,17 @@ func (t *Trojan) ListenPacketContext(ctx context.Context, metadata *C.Metadata, 
 		if err != nil {
 			return nil, fmt.Errorf("%s connect error: %w", t.addr, err)
 		}
-		defer func() {
+		defer func(c net.Conn) {
 			safeConnClose(c, err)
-		}()
+		}(c)
 	} else {
 		c, err = dialer.DialContext(ctx, "tcp", t.addr, t.Base.DialOptions(opts...)...)
 		if err != nil {
 			return nil, fmt.Errorf("%s connect error: %w", t.addr, err)
 		}
-		defer func() {
+		defer func(c net.Conn) {
 			safeConnClose(c, err)
-		}()
+		}(c)
 		tcpKeepAlive(c)
 		c, err = t.plainStream(c)
 		if err != nil {
--- a/adapter/outbound/trojan.go
+++ b/adapter/outbound/trojan.go
@@ -109,9 +109,9 @@ func (t *Trojan) DialContext(ctx context.Context, metadata *C.Metadata, opts ...
 	}
 	tcpKeepAlive(c)
 
-	defer func() {
+	defer func(c net.Conn) {
 		safeConnClose(c, err)
-	}()
+	}(c)
 
 	c, err = t.StreamConn(c, metadata)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


adapter/outbound/trojan.go
```diff
@@ -129,13 +131,17 @@ func (t *Trojan) ListenPacketContext(ctx context.Context, metadata *C.Metadata,
 		if err != nil {
 			return nil, fmt.Errorf("%s connect error: %w", t.addr, err)
 		}
-		defer safeConnClose(c, err)
+		defer func() {
+			safeConnClose(c, err)
```
Dreamacro, line 135: All safeConnClose looks like it needs to pass in the original conn

adapter/outbound/util.go
```diff
@@ -53,7 +53,7 @@ func resolveUDPAddr(network, address string) (*net.UDPAddr, error) {
 }
 
 func safeConnClose(c net.Conn, err error) {
-	if err != nil {
-		c.Close()
+	if err != nil && c != nil {
```
Dreamacro, line 56: You should make sure that the incoming conn is non-nil, if not keep the panic behavior for easy debug

adapter/outbound/util.go
```diff
@@ -53,7 +53,7 @@ func resolveUDPAddr(network, address string) (*net.UDPAddr, error) {
...
+		_ = c.Close()
```
Dreamacro, line 57: Linter allows implicit ignoring errors

Dreamacro (changes requested):

Dreamacro (approved):

Dreamacro:
Thanks a lot

xjasonlyu:
I just checked golang's `defer` behavior, considering the following code:

```go
package main

import (
	"errors"
	"fmt"
)

func main() {
	a()
}

func show(err error) {
	if err != nil {
		fmt.Printf("3st defer: %s\n", err)
	} else {
		fmt.Println("3st defer: defer not error")
	}
}

func a() {
	var err error

	defer func() {
		if err != nil {
			fmt.Printf("1st defer: %s\n", err)
		} else {
			fmt.Println("1st defer: defer not error")
		}
	}()

	defer func(err error) {
		if err != nil {
			fmt.Printf("2nd defer: %s\n", err)
		} else {
			fmt.Println("2nd defer: defer not error")
		}
	}(err)

	defer show(err)

	err = errors.New("new error")
	if err != nil {
		return
	}
}
```

Got output:

```
3st defer: defer not error
2nd defer: defer not error
1st defer: new error
```

- Demo here: https://go.dev/play/p/pRSZgLH2nDP

Therefore, I believe this code:

```go
defer func(c net.Conn) {
	safeConnClose(c, err)
}(c)
```

equals to this code:

```go
defer safeConnClose(c, err)
```

Let me know if I get it wrongly. @Dreamacro @wwqgtxx

wwqgtxx:
@xjasonlyu As shown in your test results, what we need is to maintain the behavior of `1st defer` for `err`, and fix `conn` to be original instead of streamed.

xjasonlyu:
Oh I see, so it's for getting oldest `c` and latest `err`. Thanks for your explanation!

PR #2469, opened 2022-12-20
Chore: skip dns cache for acme challenge

ffy:
Clash caches DNS records and expires with their TTL. However, this may lead to failure when using [DNS-01-Challenge](https://letsencrypt.org/docs/challenge-types/#dns-01-challenge) in most acme tools which will lookup TXT records for validation.

According to Let's Encrypt docs, such TXT record will be like `_acme-challenge.<YOUR_DOMAIN>`, so I skip cache for these queries if they fulfill both conditions.

And [a relevant issue](https://github.com/vernesong/OpenClash/issues/2280).


dns/resolver.go
```diff
@@ -4,16 +4,16 @@ import (
...
-	"strings"
-	"time"
-
 	"github.com/Dreamacro/clash/common/cache"
 	"github.com/Dreamacro/clash/component/fakeip"
 	"github.com/Dreamacro/clash/component/resolver"
 	"github.com/Dreamacro/clash/component/trie"
 	C "github.com/Dreamacro/clash/constant"
+	"github.com/Dreamacro/clash/log"
+	"math/rand"
```
Dreamacro, line 13: Please do not change the original std import sort

dns/resolver.go
```diff
@@ -134,6 +134,11 @@ func (r *Resolver) ExchangeContext(ctx context.Context, m *D.Msg) (msg *D.Msg, e
 	}
 
 	q := m.Question[0]
+	// skip dns cache for acme challenge
+	if q.Qtype == D.TypeTXT && strings.Contains(q.Name, "_acme-challenge") {
```
Dreamacro, line 138: You can filter out this msg in the `putMsgToCache`

dns/resolver.go
```diff
@@ -134,6 +134,11 @@ func (r *Resolver) ExchangeContext(ctx context.Context, m *D.Msg) (msg *D.Msg, e
...
+		log.Infoln("[DNS] dns cache ignored because of acme challenge for: %s", q.Name)
```
Dreamacro, line 139: Maybe it doesn't deserve an Info log

Dreamacro (changes requested):

ffy:
Thanks for your comments!  Changes have been pushed and I'd appreciate a re-review. @Dreamacro

dns/util.go
```diff
@@ -16,6 +17,15 @@ import (
 )
 
 func putMsgToCache(c *cache.LruCache, key string, msg *D.Msg) {
+	// skip dns cache for acme challenge
+	if len(msg.Question) != 0 {
```
Dreamacro, line 21: There is no need to check the length here, the outer layer has already been checked

ffy: Cool! This unnecessary length check has been removed.

Dreamacro (approved):

Dreamacro:
Thanks a lot

PR #2472, opened 2022-12-21
linux ipv6 redirect

embeddedlove:


embeddedlove:
image

embeddedlove:
https://github.com/Dreamacro/clash/issues/2428

listener/redir/tcp_linux.go
```diff
@@ -29,6 +29,9 @@ func parserPacket(conn net.Conn) (socks5.Addr, error) {
 
 	rc.Control(func(fd uintptr) {
 		addr, err = getorigdst(fd)
+		if err != nil {
```
Dreamacro, line 32: I have updated the `getorigdst` code, please pull and change it to the new implementation. Also instead of using `getorigdst6` for the `getorigdst` error, you should get the ip family from LocalAddr

PR #2498, opened 2023-01-17
Fix: panic on disable dns section

yaling888:


hub/route/dns.go
```diff
@@ -20,6 +20,12 @@ func dnsRouter() http.Handler {
 }
 
 func queryDNS(w http.ResponseWriter, r *http.Request) {
+	if resolver.DefaultResolver == nil {
+		render.Status(r, http.StatusInternalServerError)
+		render.JSON(w, r, newError("DNS server was disabled"))
```
Dreamacro, line 25: In fact, It is different from the DNS server, It does not return fakeip and hosts. It's helpful for debugging DNS issue

Dreamacro: maybe `DNS section is disabled`?

Dreamacro: I means `newError("DNS section is disabled")`

Dreamacro:
thanks a lot

PR #2513, opened 2023-01-30
Feature: support vmess 'zero' security

### LINKED ISSUE #2510: [Feature] 请求对vmess协议的zero加密方式进行支持
### Verify steps

- [X]  我已经在 [Issue Tracker](……/) 中找过我要提出的请求 I have searched on the [issue tracker](……/) for a related feature request. 
- [X]  我已经仔细看过 [Documentation](https://github.com/Dreamacro/clash/wiki/) 并无法自行解决问题 I have read the [documentation](https://github.com/Dreamacro/clash/wiki/) and was unable to solve the issue. 

### Description

之前有人在issue中提到过希望对vmess zero加密方式进行支持 #1390 #1699 ，但都过期了。现在使用vmess协议大都采取vmess+tls+ws的方式，采用这种方式vmess协议本身已经不需要再进行加密了，因此希望作者能添加对vmess zero加密方式的支持。

### Possible Solution

_No response_

fakeboboliu:
Fix #2510

Although it just avoids a buffer per conn, it's fairly easy, and why not make a pr?

This PR also removed an unused variable.


Dreamacro:
could you please add an test for vmess zero mode?

fakeboboliu:
> could you please add an test for vmess zero mode?

done, and also fixes vmess tests with latest(v5) v2fly/v2fly-core image.

transport/vmess/conn.go
```diff
@@ -227,6 +229,16 @@ func newConn(conn net.Conn, id *ID, dst *DstAddr, security Security, isAead bool
...
 	var reader io.Reader
 	switch security {
+	case SecurityZero:
+		security = SecurityNone
+		if !dst.UDP {
+			reader = conn
+			writer = conn
+			option = 0
+		} else {
+			reader = newChunkReader(conn)
```
Dreamacro, line 239: Is the behavior of udp inconsistent with tcp?

fakeboboliu: v2ray do treat udp in a different way

https://github.com/v2fly/v2ray-core/blob/b4069f743c3f26eee8600bc47430234da0c342a7/common/buf/io.go#L107

instead of impl a SequentialWriter, I choose just reuse Stream mode, which basically the same and does not cause other problems.

test/vmess_test.go
```diff
@@ -244,10 +256,51 @@ func TestClash_VmessWebsocketTLS(t *testing.T) {
...
+	hostCfg := &container.HostConfig{
+		PortBindings: defaultPortBindings,
+		Binds: []string{
+			fmt.Sprintf("%s:/etc/v2ray/config.json", C.Path.Resolve("vmess-ws-tls-zero.json")),
+			fmt.Sprintf("%s:/etc/ssl/v2ray/fullchain.pem", C.Path.Resolve("example.org.pem")),
+			fmt.Sprintf("%s:/etc/ssl/v2ray/privkey.pem", C.Path.Resolve("example.org-key.pem")),
+		},
+	}
+
+	id, err := startContainer(cfg, hostCfg, "vmess-ws")
```
Dreamacro, line 275: should be "vmess-ws-tls-zero"?

fakeboboliu: well, all containers in vmess_test.go are using the same container name, so I'm in.

Dreamacro: you can clean up it by the way 😆

transport/vmess/conn.go
fakeboboliu: there may be some mistakes, v2ray will run an AEAD without E mechanism for UDP requests while using None.

https://github.com/v2fly/v2ray-core/blob/b4069f743c3f26eee8600bc47430234da0c342a7/proxy/vmess/encoding/server.go#L325

So clash's current None is somehow broken.

Dreamacro: I guess it's because udp is not streaming and must have some field like `length`.

Dreamacro:
thanks

PR #2597, opened 2023-03-09
Feature: add support for search domains

jeff-an:
Environments like `docker` and `kubernetes` often have default DNS resolvers that contain [search domains](https://en.wikipedia.org/wiki/Search_domain) that make service discovery easier. For example, within a single Kubernetes namespace, I can reach a Postgres pod at `http://postgres` without needing to specify the fully qualified name. For example:

```
$ cat /etc/resolv.conf
nameserver 172.10.0.10
search my-namespace.svc.cluster.local svc.cluster.local cluster.local us-west-2.compute.internal
options ndots:5
```

Unfortunately, there is no way to preserve this configuration when using `fake-ip` in `clash` today. If the `search` field is left in the `/etc/resolv.conf` and the `resolv.conf` is changed so that all DNS queries go to clash's resolver at `0.0.0.0:53`, **all** queries will have the search domain automatically appended, even if it is not needed. For example:

```
WARN[0001] [TCP] dial http (match Match/) 10.50.0.101:48866 --> postgres.my-namespace.svc.cluster.local.my-namespace.svc.cluster.local:5432 error: 503 Service Unavailable
```

This PR allows users to configure search domains they would like to add to queries if the first query does not succeed. This mimics the default linux behavior. Queries are done in parallel and thus the performance hit should be minimal. Finally, since search domains are only applied during the true DNS resolution flow, the fake-ip pool is not affected.

### DIFF
--- a/dns/resolver.go
+++ b/dns/resolver.go
@@ -298,7 +298,7 @@ func (r *Resolver) lookupIP(ctx context.Context, host string, dnsType uint16) ([
 		return nil, resolver.ErrIPNotFound
 	}
 
-	// query DNS servers in parallel for each search domain
+	// query provided search domains serially
 	for _, domain := range r.searchDomains {
 		q := &D.Msg{}
 		q.SetQuestion(D.Fqdn(fmt.Sprintf("%s.%s", host, domain)), dnsType)
--- a/dns/resolver.go
+++ b/dns/resolver.go
@@ -7,6 +7,7 @@ import (
 	"math/rand"
 	"net"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Dreamacro/clash/common/cache"
@@ -39,6 +40,7 @@ type Resolver struct {
 	group                 singleflight.Group
 	lruCache              *cache.LruCache
 	policy                *trie.DomainTrie
+	searchDomains         []string
 }
 
 // LookupIP request with TypeA and TypeAAAA, priority return TypeA
@@ -291,10 +293,49 @@ func (r *Resolver) lookupIP(ctx context.Context, host string, dnsType uint16) ([
 	}
 
 	ips := msgToIP(msg)
-	if len(ips) == 0 {
+	if len(ips) != 0 {
+		return ips, nil
+	}
+	if len(ips) == 0 && len(r.searchDomains) == 0 {
 		return nil, resolver.ErrIPNotFound
 	}
-	return ips, nil
+
+	// query search domains in parallel, taking the first result returned
+	wg := sync.WaitGroup{}
+	doneCh := make(chan struct{})
+	resultCh := make(chan []net.IP, len(r.searchDomains))
+	resultReceivedCh := make(chan struct{})
+	for _, domain := range r.searchDomains {
+		wg.Add(1)
+		go (func(d string) {
+			defer wg.Done()
+			q := &D.Msg{}
+			q.SetQuestion(D.Fqdn(fmt.Sprintf("%s.%s", host, d)), dnsType)
+			msg, err := r.Exchange(q)
+			if err != nil {
+				return
+			}
+			ips := msgToIP(msg)
+			if len(ips) != 0 {
+				resultCh <- ips
+				<-resultReceivedCh
+			}
+		})(domain)
+	}
+	go (func() {
+		wg.Wait()
+		close(doneCh)
+	})()
+
+	select {
+	case ips = <-resultCh:
+		close(resultReceivedCh)
+		return ips, nil
+	case <-doneCh:
+		break
+	}
+
+	return nil, resolver.ErrIPNotFound
 }
 
 func (r *Resolver) msgToDomain(msg *D.Msg) string {
@@ -336,6 +377,7 @@ type Config struct {
 	Pool           *fakeip.Pool
 	Hosts          *trie.DomainTrie
 	Policy         map[string]NameServer
+	SearchDomains  []string
 }
 
 func NewResolver(config Config) *Resolver {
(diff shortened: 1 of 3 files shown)


Dreamacro:
Thanks for the detailed description and the PR.

As far as I know, there may be only one dns server that needs to search domain, is it too expensive to do domain search for every dns server in each nameserver? The current PR only requests search domain for A/AAAA records, do other record types need to be requested as well (I haven't had time to look through the RFC yet)?

A simple helper is currently available in the repository for sending msg in parallel, see
https://github.com/Dreamacro/clash/blob/fbf2f265160819116dabfb2b2427d5c58aa11767/dns/util.go#L117-L143

BTW, recently I've been investigating and wanting to add a Multicast DNS client, which also includes search domains. It was initially designed as a separate client configuration. 

```yaml
dns:
    nameserver:
      - mdns://local
```

But it seems better as a field of dns:
```yaml
dns:
    mdns: [.local]
```

If this pr is merged, it seems to be better:
```yaml
dns:
    mdns: true
    search-domains: [.local]
```

Looking forward to hearing your opinion.

jeff-an:
Thanks for the interesting questions! I will switch to using the helper function.

1. Regarding if we should search only one DNS server instead of all of them: I think this raises a good point. A common situation might be that people have one public nameserver (e.g. 8.8.8.8) and another private one that needs to handle search domains. (e.g. 10.x.x.x). So maybe, the configuration of search domains should be specific to each nameserver? e.g.

``` yaml
dns:
  nameservers:
  - addr: 172.10.0.10
     search-domains: ['my-namespace.svc.cluster.local', 'svc.cluster.local']
  - addr: 8.8.8.8
```
For simplicity and backwards compatibility, perhaps we can continue to accept simple strings under nameservers if we go this route.

2. Other records: I suppose it is possible, but I have not heard of any real life use cases where other records are requested with a search domain. The only thing I can maybe think of is an MX record for a private network email client? And `clash` wouldn't work with those records anyways right?

3. Interesting, I think it might fit well with the per-nameserver configuration too, since users might want to use multi-cast along with regular DNS servers.
```yaml
dns:
  nameservers:
  - addr: mdns://local
     search-domains: ['.local']
   ...
```

Dreamacro:
The problem with putting it in the nameserver is that it complicates the configuration, but in the case of a small number of nameservers, it is impossible to use it globally as a `dns` field.

> And clash wouldn't work with those records anyways right?

The real problem is this. Yes, clash now contains other dns records, mainly used to provide a generic dns server with `listen` field. I don't really know if search domains are used for other types of records, as far as I know, there is another common record `PTR` for search domains, which is used for reverse domain name lookup, and this is also used in Multicast DNS

jeff-an:
> The problem with putting it in the nameserver is that it complicates the configuration, but in the case of a small number of nameservers, it is impossible to use it globally as a dns field.

I agree it is a complication and making it backwards compatible would be difficult as well. How about we stick with the original configuration format for now, and we can add a disclaimer that this may affect performance? In my own tests however I have not noticed anything major.

> as far as I know, there is another common record PTR for search domains

Ah I did not think of this use case yet. Seems like clash would handle this well? It already has a mapping from IP address to host name, and the host name would include the search domain. Clash can then perform the reverse DNS lookup with the real nameservers. In any case, a little out of scope for this PR I think? 😅

Dreamacro (changes requested):
It seems that there is only a need for helper replacement. Other than that, I have no problem.

dns/resolver.go
```diff
@@ -291,10 +293,49 @@ func (r *Resolver) lookupIP(ctx context.Context, host string, dnsType uint16) ([
 	}
 
 	ips := msgToIP(msg)
-	if len(ips) == 0 {
+	if len(ips) != 0 {
+		return ips, nil
+	}
+	if len(ips) == 0 && len(r.searchDomains) == 0 {
```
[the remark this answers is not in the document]
jeff-an: good call

jeff-an:
> It seems that there is only a need for helper replacement. Other than that, I have no problem.

i switched the code to use `r.ExchangeContext` which calls the helper function you linked under the hood. I also changed another call to use `ExchangeContext` as well - since we have a `ctx` in `lookupIP` we might as well use it right 😃

dns/resolver.go
```diff
@@ -285,16 +286,33 @@ func (r *Resolver) lookupIP(ctx context.Context, host string, dnsType uint16) ([
...
 	ips := msgToIP(msg)
-	if len(ips) == 0 {
+	if len(ips) != 0 {
+		return ips, nil
+	} else if len(r.searchDomains) == 0 {
 		return nil, resolver.ErrIPNotFound
 	}
-	return ips, nil
+
+	// query DNS servers in parallel for each search domain
```
Dreamacro, line 301: It looks like there is no longer a parallel here, right?

Dreamacro (changes requested):
Thanks for finding the glitch in `Exchange`.

I'm also confused about the github wiki moderation permissions because I found other repository wikis that anyone can edit.

dns/resolver.go
jeff-an:
```
func batchExchange(ctx context.Context, clients []dnsClient, m *D.Msg) (msg *D.Msg, err error) {
	fast, ctx := picker.WithContext(ctx)
	for _, client := range clients {
		r := client
		fast.Go(func() (any, error) {
			m, err := r.ExchangeContext(ctx, m)
			if err != nil {
				return nil, err
			} else if m.Rcode == D.RcodeServerFailure || m.Rcode == D.RcodeRefused {
				return nil, errors.New("server failure")
			}
			return m, nil
		})
	}

	elm := fast.Wait()
```

^ `fast` seems to invoke the queries in parallel

jeff-an: yes that's correct. the new commit is ok right? are you just referring to the comment?

Dreamacro (approved):

Dreamacro:
Thanks a lot, would you mind if this commit cherry-pick to premium?

ingjieye:
Hi @jeff-an @Dreamacro, after this merge, how should I update the config to use the mDNS feature? The wiki seems to not update accordingly.

Issue #2524, opened 2023-02-04
[Bug] 开启 IPV6 后内核崩溃 panic: runtime error: index out of range [0] with length 0


iKeqing:
### Verify steps

- [X]  确保你使用的是**本仓库**最新的的 clash 或 clash premium 版本 Ensure you are using the latest version of Clash or Clash Premium from **this repository**. 
- [X]  如果你可以自己 debug 并解决的话，提交 PR 吧 Is this something you can **debug and fix**? Send a pull request! Bug fixes and documentation fixes are welcome. 
- [X]  我已经在 [Issue Tracker](……/) 中找过我要提出的问题 I have searched on the [issue tracker](……/) for a related issue. 
- [X]  我已经使用 dev 分支版本测试过，问题依旧存在 I have tested using the dev branch, and the issue still exists. 
- [X]  我已经仔细看过 [Documentation](https://github.com/Dreamacro/clash/wiki/) 并无法自行解决问题 I have read the [documentation](https://github.com/Dreamacro/clash/wiki/) and was unable to solve the issue. 
- [X]  这是 Clash 核心的问题，并非我所使用的 Clash 衍生版本（如 OpenClash、KoolClash 等）的特定问题 This is an issue of the Clash core *per se*, not to the derivatives of Clash, like OpenClash or KoolClash. 

### Clash version

Premium 2023.01.29

### What OS are you seeing the problem on?

Linux

### Clash config

```yaml
---
port: 7890
socks-port: 7891
redir-port: 7892
tproxy-port: 7895
mixed-port: 7893
allow-lan: true
bind-address: "*"
mode: rule
log-level: info
ipv6: true
hosts:
  services.googleapis.cn: 216.58.200.67
  www.google.cn: 216.58.200.67
external-controller: 0.0.0.0:9090
clash-for-android:
  append-system-dns: false
profile:
  tracing: true
dns:
  enable: true
  ipv6: true
  listen: 0.0.0.0:7874
  enhanced-mode: fake-ip
  fake-ip-range: 198.18.0.1/16
  use-hosts: true
  default-nameserver:
  - "[2402:4e00::]:53"
  - 119.29.29.29
  nameserver:
  - "[2402:4e00::]:53"
  - 119.29.29.29
```

### Clash log

```shell
github.com/miekg/dns@v1.1.50/server.go:533 +0x3f0
created by github.com/miekg/dns.(*Server).serveUDP
[...]

PR #2634, opened 2023-03-21
Fix: potential panic in putMsgToCache

lujjjh:
When the upstream DNS server returns a message that contains no questions (i.e. QDCOUNT == 0), `putMsgToCache` will trigger an out of range panic.

Issue: #2524
Comment: https://github.com/Dreamacro/clash/issues/2524#issuecomment-1477477601


Dreamacro:
Makes sense, but under what circumstances does the dns server not bring the Question when it returns?

Dreamacro:
found https://github.com/Dreamacro/clash/issues/2524#issuecomment-1477477601. Got it

PR #2643, opened 2023-03-26
Fix: return pooled buffer when simple-obfs tls read error

bdbai:


transport/simple-obfs/tls.go
```diff
@@ -28,6 +28,7 @@ func (to *TLSObfs) read(b []byte, discardN int) (int, error) {
 	buf := pool.Get(discardN)
 	_, err := io.ReadFull(to.Conn, buf)
 	if err != nil {
+		pool.Put(buf)
```
Dreamacro, line 31: maybe just put this before `if err != nil` ?

bdbai: I agree this can be simplified, but it seems to break the convention that `if err != nil` immediately follows the statement where `err` comes. wdyt?

Dreamacro: If golangci-lint doesn't complain, I don't think it's a problem.

bdbai: Fixed.

Dreamacro:
cool, thanks

PR #2647, opened 2023-03-28
Chore: Allow http to set custom headers

RommHui:


RommHui:
An exmaple configuration can be:
`
proxies:
    - name: example
      type: http
      server: 127.0.0.1
      port: 443
      headers:
        User-Agent: chrome
        X-Forwarded-For: 8.8.8.8
`

Dreamacro:
Can you explain what specific custom header solves your problem?

RommHui:
@Dreamacro Some http proxies need to verify certain specific request headers, but the original http proxy of Clash does not support custom request headers, which causes the http proxy to fail to operate normally

Dreamacro:
@RommHui For example? I want to see a practical example.

adapter/outbound/http.go
```diff
@@ -134,6 +136,11 @@ func NewHttp(option HttpOption) *Http {
 		}
 	}
 
+	headers := make(http.Header)
```
Dreamacro, line 139: perfer `http.Header{}` in this repo

Dreamacro (changes requested):

adapter/outbound/http.go
RommHui: `make(http.Header)` has been changed to `http.Header{}`

Dreamacro:
Thanks for contributing

PR #2659, opened 2023-04-04
Fix: should always drop packet when handle UDP session

yaling888:
should always drop packet when handle UDP session


tunnel/tunnel.go
```diff
@@ -225,7 +232,7 @@ func handleUDPConn(packet *inbound.PacketAdapter) {
 		if loaded {
```
Dreamacro, line 232: It looks like you can `defer packet.Drop()` directly in this goroutine. And remove `defer packet.Drop()` in handleUDPToRemote

Dreamacro:
You may have misunderstood. `handleUDPConn` have a goroutine, so put `defer packet.Drop()` is not safe. Should put it on goroutine

Dreamacro:
Thanks for contributing

PR #2672, opened 2023-04-10
Feature: reuse tcp connection in HTTP vehicle

kamingchan:
- use global http client to reuse tcp connection between different HTTP vehicles
- set `ForceAttemptHTTP2` to true to enable HTTP2


Dreamacro:
Reusing connections is good, but it looks like making provider go through a specific proxy becomes impossible

kamingchan:
Making provider go through a specific proxy? I couldn't find such a feature in code base. Is that only in premium core or just in the roadmap? 😄

PR #2684, opened 2023-04-18
Feature: support basic authentication for DoH

damnever:


Dreamacro:
Some detailed explanation of pr is needed, such as where the DoH requires basic auth

damnever:
@Dreamacro There are some DoH servers doesn't want to share themself.

> The DoH protocol design allows applications to fully leverage the HTTP ecosystem, including features that are not enumerated [here](https://www.rfc-editor.org/rfc/rfc8484).

Dreamacro:
I'm more interested in knowing which real-world DoH needs Basic Auth

damnever:
I am hosting such a DoH server to do some traffic management which meets my need, and I believe there are many private DoH servers as well.

So What do you think?

Dreamacro:
I don't have any other opinions. I just want to know the scene that everyone uses.

Have you verified that it works?

damnever:
yes, `https://user:pass@dohserver` works for me

Dreamacro:
cool, thanks

PR #2693, opened 2023-04-21
Support IPSET rule

major1201:
This PR makes clash support IPSet rules.

Example config fragment:

```yaml
rules:
  - IPSET,corp,mycompanyproxy
  - IPSET,home,myproxy
  - MATCH,default
```


rule/ipset.go
```diff
@@ -0,0 +1,43 @@
+//go:build !linux
```
Dreamacro, line 1: I think the ipset-related code should be moved to `component` so there is no need to split many files under `rule`.

rule/ipset.go
```diff
@@ -0,0 +1,43 @@
...
+func (f *IPSet) Adapter() string {
+	return f.adapter
+}
+
+func (f *IPSet) Payload() string {
+	return f.name
+}
+
+func (f *IPSet) ShouldResolveIP() bool {
+	return false
```
Dreamacro, line 31: He should also belong to the `no-resolve` category of rules

rule/ipset_linux.go
```diff
@@ -0,0 +1,66 @@
...
+
+package rules
+
+import (
+	C "github.com/Dreamacro/clash/constant"
+	"github.com/Dreamacro/clash/log"
+	"github.com/vishvananda/netlink"
+)
+
+func init() {
```
Dreamacro, line 11: Why do we need these prints?

rule/ipset_linux.go
```diff
@@ -0,0 +1,66 @@
...
+	return false
+}
+
+func (f *IPSet) ShouldFindProcess() bool {
+	return false
+}
+
+func NewIPSet(name string, adapter string) *IPSet {
+	_, err := netlink.IpsetList(name)
+	if err != nil {
```
Dreamacro, line 58: why not just return an error?

Dreamacro (changes requested):

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/samber/lo v1.38.1
 	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.2
+	github.com/vishvananda/netlink v1.2.1-beta.2.0.20230420174744-55c8b9515a01
```
Dreamacro, line 18: Maybe update the dependency to the main instead of the tag version

rule/ipset_linux.go
major1201: Got it, I'll keep the logs clean, the check can also be done in `NewIPSet()`

go.mod
major1201: The `netlink.IpsetTest()` has been merged recently to the main branch in [github.com/vishvananda/netlink](github.com/vishvananda/netlink) and the last release date is Jan 2020, so I'm not sure when the next version comes.

Dreamacro: you can replace `v1.2.1-beta.2.0.20230420174744-55c8b9515a01` to `main` and run `go mod tidy`

component/ipset/ipset_linux.go
```diff
@@ -0,0 +1,22 @@
...
+
+// Test whether the ip is in the set or not
+func Test(setName string, ip net.IP) (bool, error) {
+	return netlink.IpsetTest(setName, &netlink.IPSetEntry{
+		IP: ip,
+	})
+}
+
+// Check dumps a specific ipset to check if the set exists or not
+func Check(setName string) error {
```
Dreamacro, line 19: Maybe a better name would be `Exist`?

go.mod
major1201: I did, but after `go mod tidy`, it recovered to `v1.2.1-beta.2.0.20230420174744-55c8b9515a01` which is the [go mod pseudo-version](https://go.dev/doc/modules/version-numbers) of the latest commit of the main branch generated by go (ref: [go mod file reference](https://go.dev/doc/modules/gomod-ref)).

component/ipset/ipset_linux.go
major1201: Sorry, the comment of the func may be confusing, I use `ipset list` to check the set can be used normally. If `err != nil` the reason can be the set not exists, but it can also be other reasons. Maybe let's use `TryList`?

go.mod
Dreamacro: Oh, this is the latest, I forgot about that

component/ipset/ipset_linux.go
Dreamacro: what about `Verify`

Dreamacro (approved):

Dreamacro:
would you mind cherry-pick it to premium?

Dreamacro:
BTW, when DstIP is nil, What does `IpsetTest` return?

major1201:
> would you mind cherry-pick it to premium?

Sure I can do that.

major1201:
> BTW, when DstIP is nil, What does `IpsetTest` return?

It'll return `false, nil`

PR #2696, opened 2023-04-23
Assume 403 status code as timeout in urlTest to filter blocked proxies.

kingofnull:
When a webserver block or black list a proxy (like Spotify or developer.android.com) urlTest return a valid ping to to target url while it has been blocked! Then I just add a condition to behave 403 response code as timeout , now we can filter out blacklisted proxies by urlTest.


Aobasama:
It seems that my patch can cover your need. https://github.com/Dreamacro/clash/pull/2626

Dreamacro:
Duplicate of #2626

PR #2700, opened 2023-04-25
Fix: socks4 server handshake

yaling888:
Wrong order.


Dreamacro:
thanks for catch this 🤣

PR #2706, opened 2023-04-26
Fix: Adjust DNS TTL values based on minimum value

KaitoHH:
## Problem
The following code uses the TTL of the very first record, as the whole TTL of the response,
https://github.com/Dreamacro/clash/blob/df61a586c9df496c2ab71790534b0cab5d9e47ba/dns/util.go#L28-L33

and these codes further use the previous TTL as the TTL of each entry in answers.
https://github.com/Dreamacro/clash/blob/df61a586c9df496c2ab71790534b0cab5d9e47ba/dns/util.go#L42-L55

This will cause issue when the first answer in the DNS message is greater than the later record.

For example, in the first DNS query
```bash
╰─❯ dog epdg.epc.mnc260.mcc310.pub.3gppnetwork.org A @openwrt.lan
CNAME epdg.epc.mnc260.mcc310.pub.3gppnetwork.org.     5h49m11s   "epdg.epc.geo.mnc260.mcc310.pub.3gppnetwork.org."
    A epdg.epc.geo.mnc260.mcc310.pub.3gppnetwork.org.      17s   208.54.39.163
```

And the second cached query will cause the answer of A type to have a much greater TTL value, which should be expected to expire after 17 seconds after the first query.
```bash
╰─❯ dog epdg.epc.mnc260.mcc310.pub.3gppnetwork.org A @openwrt.lan
CNAME epdg.epc.mnc260.mcc310.pub.3gppnetwork.org.     5h49m07s   "epdg.epc.geo.mnc260.mcc310.pub.3gppnetwork.org."
    A epdg.epc.geo.mnc260.mcc310.pub.3gppnetwork.org. 5h49m07s   208.54.39.163
```

## Resolution
To resolve the issues, this commit adds modify `updateTTL` function that adjusts the TTL values of DNS records based on the minimum TTL value found in the records list instead of the first record. When updating the TTL it calculates the passed delta time and subtracts it to all cached TTL values.  This ensures consistency in the cache expiry time for all records to prevent caching issues.

### DIFF
--- a/dns/util.go
+++ b/dns/util.go
@@ -29,7 +29,7 @@ func updateTTL(records []D.RR, ttl uint32) {
 	}
 	delta := minimalTTL(records) - ttl
 	for i := range records {
-		records[i].Header().Ttl -= delta
+		records[i].Header().Ttl = lo.Clamp(records[i].Header().Ttl-delta, 1, records[i].Header().Ttl)
 	}
 }
 
--- a/dns/util.go
+++ b/dns/util.go
@@ -57,6 +57,20 @@ func putMsgToCache(c *cache.LruCache, key string, q D.Question, msg *D.Msg) {
 }
 
 func setMsgTTL(msg *D.Msg, ttl uint32) {
+	for _, answer := range msg.Answer {
+		answer.Header().Ttl = ttl
+	}
+
+	for _, ns := range msg.Ns {
+		ns.Header().Ttl = ttl
+	}
+
+	for _, extra := range msg.Extra {
+		extra.Header().Ttl = ttl
+	}
+}
+
+func updateMsgTTL(msg *D.Msg, ttl uint32) {
 	updateTTL(msg.Answer, ttl)
 	updateTTL(msg.Ns, ttl)
 	updateTTL(msg.Extra, ttl)
--- a/dns/util.go
+++ b/dns/util.go
@@ -16,6 +16,26 @@ import (
 	D "github.com/miekg/dns"
 )
 
+func minimalTTL(records []D.RR) uint32 {
+	ttl := records[0].Header().Ttl
+	for i := range records {
+		if records[i].Header().Ttl < ttl {
+			ttl = records[i].Header().Ttl
+		}
+	}
+	return ttl
+}
+
+func updateTTL(records []D.RR, ttl uint32) {
+	if len(records) == 0 {
+		return
+	}
+	delta := minimalTTL(records) - ttl
+	for i := range records {
+		records[i].Header().Ttl -= delta
+	}
+}
+
 func putMsgToCache(c *cache.LruCache, key string, q D.Question, msg *D.Msg) {
 	// skip dns cache for acme challenge
 	if q.Qtype == D.TypeTXT && strings.HasPrefix(q.Name, "_acme-challenge.") {
@@ -26,11 +46,11 @@ func putMsgToCache(c *cache.LruCache, key string, q D.Question, msg *D.Msg) {
 	var ttl uint32
 	switch {
 	case len(msg.Answer) != 0:
-		ttl = msg.Answer[0].Header().Ttl
+		ttl = minimalTTL(msg.Answer)
 	case len(msg.Ns) != 0:
-		ttl = msg.Ns[0].Header().Ttl
+		ttl = minimalTTL(msg.Ns)
 	case len(msg.Extra) != 0:
-		ttl = msg.Extra[0].Header().Ttl
+		ttl = minimalTTL(msg.Extra)
 	default:
 		log.Debugln("[DNS] response msg empty: %#v", msg)
 		return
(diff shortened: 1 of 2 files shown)


dns/util.go
```diff
@@ -16,6 +16,26 @@ import (
 	D "github.com/miekg/dns"
 )
 
+func minimalTTL(records []D.RR) uint32 {
+	ttl := records[0].Header().Ttl
```
Dreamacro, line 20: Wouldn't it be easier to use https://github.com/samber/lo#minby?

dns/util.go
```diff
@@ -40,17 +60,9 @@ func putMsgToCache(c *cache.LruCache, key string, q D.Question, msg *D.Msg) {
...
-	}
-
-	for _, ns := range msg.Ns {
-		ns.Header().Ttl = ttl
-	}
-
-	for _, extra := range msg.Extra {
-		extra.Header().Ttl = ttl
-	}
+	updateTTL(msg.Answer, ttl)
```
Dreamacro, line 63: The ttl here is already minimalTTL delta.

Dreamacro (changes requested):

dns/util.go
KaitoHH: Here ttl is the value that each record should be set, not the delta value, so we should first calculate the time elapsed (delta = record.TTL - ttl), and then do the subtraction.
Another way of fixing this is to refactor the updateTTL function by passing the delta parameter, but that will be a bit more complex.

Dreamacro: `setMsgTTL(msg.Msg, uint32(time.Until(expireTime).Seconds()))` It's calculated by the expiration time of cache. It's delta.

KaitoHH: I made a mistake in my previous explanation. Let me clarify, at first we should set TTL to be `uint32(time.Until(expireTime).Seconds())`, but this will cause each record to have the same TTL value. 

Let's suppose 5s passed by the time of the first query, so instead of
```bash
╰─❯ dog epdg.epc.mnc260.mcc310.pub.3gppnetwork.org A @openwrt.lan
CNAME epdg.epc.mnc260.mcc310.pub.3gppnetwork.org.     12s  "epdg.epc.geo.mnc260.mcc310.pub.3gppnetwork.org."
    A epdg.epc.geo.mnc260.mcc310.pub.3gppnetwork.org.      12s   208.54.39.163
```

We subtract the common value from each TTL record, so the result will become
 ```bash
╰─❯ dog epdg.epc.mnc260.mcc310.pub.3gppnetwork.org A @openwrt.lan
CNAME epdg.epc.mnc260.mcc310.pub.3gppnetwork.org.     5h49m6s   "epdg.epc.geo.mnc260.mcc310.pub.3gppnetwork.org."
    A epdg.epc.geo.mnc260.mcc310.pub.3gppnetwork.org.      12s   208.54.39.163
```

Comment (Dreamacro) [reply dns/util.go:60]: In fact the logic of setMsgTTL now sets an absolute value, and there are several other places where it is forced to 1. For these places the current change breaks this behavior

Comment (KaitoHH) [reply dns/util.go:60]: How about splitting it into two functions so that the fakeip logic is not affected?

Comment (Dreamacro) [reply dns/util.go:60]: It sounds good to use a new function.

File: dns/util.go [hunk 3]
```diff
@@ -12,10 +12,27 @@ import (
...
+	return lo.MinBy(records, func(r1 D.RR, r2 D.RR) bool {
+		return r1.Header().Ttl < r2.Header().Ttl
+	}).Header().Ttl
+}
+
+func updateTTL(records []D.RR, ttl uint32) {
+	if len(records) == 0 {
+		return
+	}
+	delta := minimalTTL(records) - ttl
```
Dreamacro, line 30: In order to prevent possible 0 and overflow problems, should we need to use `lo.Clamp`? seems we don't need `len(records)` guard?

BTW, run `make lint-fix` to fix some lint issue

KaitoHH: `len(records)` is essential since `minimalTTL()` cannot receive an empty slice. Otherwise, `Header()` on empty slice will cause nil pointer panic.

dns/util.go
```diff
@@ -14,8 +14,25 @@ import (
...
+	for i := range records {
+		records[i].Header().Ttl = lo.Clamp(records[i].Header().Ttl-delta, 1, records[i].Header().Ttl)
```
Dreamacro, line 32: https://github.com/samber/lo#clamp

the first and third parameters is lower and upper bounds.

Dreamacro (changes requested):

dns/util.go
Dreamacro: sorry, I misread it.

Dreamacro (approved):

Dreamacro:
Thanks for the patient modifications, do you mind if I port a similar logic to premium?

KaitoHH:
> Thanks for the patient modifications, do you mind if I port a similar logic to premium?

No problem.

PR #2723, opened 2023-05-10
docs: new documentation site

BirkhoffLee:
This commit adds a VitePress build to the main repository, aiming to ditch GitHub Wiki. Moving further, we're going to host our own documentation site eithor on GitHub Pages or something alike.


BirkhoffLee:
To test out: run `pnpm install` and `pnpm run docs:dev`. Other commands are resided in `/packages.json`

BirkhoffLee:
Deployment: https://vitepress.dev/guide/deploy#github-pages

> In your theme config file, docs/.vitepress/config.js, set the base property to the name of your GitHub repository. If you plan to deploy your site to https://foo.github.io/bar/, then you should set base to '/bar/'. It should always start and end with a slash.

If we're going to host it at https://dreamacro.github.io/clash, we need to modify `docs/.vitepress/config.js`, otherwise this PR is ready to merge

Dreamacro:
It depends on npm instead of pnpm. Perhaps we could do without the lock file, which is large and reduces a lot of changes to the git file since it will be updated frequently

Dreamacro:
It depends on npm instead of pnpm. Perhaps we could do without the lock file, which is large and reduces a lot of changes to the git file since it will be updated frequently

BirkhoffLee:
> It depends on npm instead of pnpm. Perhaps we could do without the lock file, which is large and reduces a lot of changes to the git file since it will be updated frequently

If you mean lockfile changes often due to the use of `pnpm install`, it indeed introduces frequent changes in git history. However dropping the lockfile can lead to build inconsistencies - hence the default preference for CI environments being [dependent on the lockfiles](https://pnpm.io/cli/install#--frozen-lockfile).

BirkhoffLee:
> It depends on npm instead of pnpm. Perhaps we could do without the lock file, which is large and reduces a lot of changes to the git file since it will be updated frequently

If you mean lockfile changes often due to the use of `pnpm install`, it indeed introduces frequent changes in git history. However dropping the lockfile can lead to build inconsistencies - hence the default preference for CI environments being [dependent on the lockfiles](https://pnpm.io/cli/install#--frozen-lockfile).

BirkhoffLee:
> Inconsistencies are caused by long periods of non-updating

Inconsistencies are caused exactly by not using lockfiles - as time goes by, the same manifests described in `package.json` can result in largely different `node_modules` (dependencies), and we will have inconsistent documentation site generated unexpectedly, and this doesn't even take long - a day or less might just be long enough for that to happen.

> so if the project is still active, there is no need to use lockfile

Let's say we drop the use of lockfiles anyway, it will still bring degraded performance where CI spends more time on package resolution.

Refer to the following on the descriptions of lockfiles:

1. https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#description
2. https://pnpm.io/git#lockfiles

Also, consider the fact that `package-lock.json` *does not* change so often in our scenario - 

1. Most of the people who want to contribute to the docs would only propose changes to `.md` files, directly on the GitHub web editor.
2. Even if the active maintainers are updating the documentations frequently, they wouldn't have run `pnpm install` so often - `node_modules` is already present, there's no reason to do that when you're updating Markdown files.

If we want to prevent lockfiles being updated, we could enforce that people always use `pnpm install --frozen-lockfile` when initially fetching the node dependencies.

Dreamacro:
> Inconsistencies are caused exactly by not using lockfiles.

I have several years of front-end development experience. These cases are rare and can be detected in time.

I don't know how to count all the lines that have changed in a file in git, but if you know how to do that, you can count how many lockfile files have changed here at https://github.com/Dreamacro/clash-dashboard/commits/master (even before it was yarn. lock)

For me this is not very acceptable in a go project, I update all dependencies at each release, including of course this documentation

BirkhoffLee:
Either way, it's your repo! I removed lockfiles and updated CI instructions. If the documentations looks good, you can merge now, otherwise you can edit directly.

PR #2731, opened 2023-05-12
Fix: incorrect UDP PacketConn reused

Hackerl:
An existing PacketConn should be identified by (sip, sport, dip, dport), because the same local UDP port can send data to any address, if only (sip, sport) is used as the key, it will result in using the same PacketConn to send data to different targets, thus throw "ErrUDPRemoteAddrMismatch" error.

### DIFF
--- a/tunnel/tunnel.go
+++ b/tunnel/tunnel.go
@@ -207,12 +207,15 @@ func handleUDPConn(packet *inbound.PacketAdapter) {
 		metadata.DstIP = ips[0]
 	}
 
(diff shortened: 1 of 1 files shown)


Dreamacro:
4-tuple should not be used as keys, clash proxy PacketConn instead of UDPConn. Vmess has inherent problems, It doesn't support full conn udp nat. Each connection only support each dstip:dport

Dreamacro:
If you need to proxy udp packet, you should keep away from Vmess.

PR #2765, opened 2023-05-24
Fix: support shadowsocks obfs http proxy

pengchongfu:
### Summary

```
proxies:
  - name: "ss-cloudflare"
    type: ss
    server: www.my-domain.com
    port: 80
    cipher: chacha20-ietf-poly1305
    password: "12345678"
    plugin: obfs
    plugin-opts:
      mode: http
      host: www.my-domain.com
```
`www.my-domain.com` is proxied by cloudflare so the traffic is `clash` -> `cloudflare` -> `shadowsocks server`

The shadowsocks configuration failed to work with clash, but it works well with iOS Potatso Lite(it also works well with clash by disabling cloudflare proxy)
After some investigations I found the root cause

1. clash will send 3 packets once connected to cloudflare: iv packet, socks addr packet and the first data packet
1. cloudflare would proxy the iv packet to shadowsocks server(See the screenshot below) but ignored socks addr packet and the first data packet because at that time the connnection didn't upgrade to websocket
1. send all the 3 packets together, it works well

Screenshot from 2023-05-24 14-34-42-redacted_dot_app

### Test

- configuration with cloudflare proxy works well
- configuration without cloudflare proxy works well

### DIFF
--- a/transport/simple-obfs/http.go
+++ b/transport/simple-obfs/http.go
@@ -16,12 +16,13 @@ import (
 // HTTPObfs is shadowsocks http simple-obfs implementation
 type HTTPObfs struct {
 	net.Conn
-	host          string
-	port          string
-	buf           []byte
-	offset        int
-	firstRequest  bool
-	firstResponse bool
+	host                string
+	port                string
+	buf                 []byte
+	offset              int
+	firstRequest        bool
+	firstRequestBufList [][]byte
+	firstResponse       bool
 }
 
 func (ho *HTTPObfs) Read(b []byte) (int, error) {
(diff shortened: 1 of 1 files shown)


Dreamacro:
This has a fatal problem: not all TCP connections are made by the client sending data to the server first (although most are), so when the server sends data first, the connection hangs

pengchongfu:
> This has a fatal problem: not all TCP connections are made by the client sending data to the server first (although most are), so when the server sends data first, the connection hangs

Thanks for your reply! You are right that this change would cause problem if the server sends data first.

I read some code of simple-obfs and shadowsocks-libev, and I found that they don't follow the rigth websocket protocol, I believe it is the root cause why cloudflare proxy doesn't work well
1. shadowsocks-libev send first packet(iv + first encrypted data) to simple-obfs when first data packet from remote server arrives
2. and then simple-obfs send http response to shandowsocks client

I think to support websocket proxy like cloudflare, a good design should be
- client sends http request and waits for the response
- server sends the http response to establish the websocket connection
- after the websocket handshake finished, the client or server start to send first data packet

Obviously, the design would take more time to wait for websocket handshake. And since simple-obfs is deprecated, taking some other protocols or plugins is better
I would close this PR

Dreamacro:
shadowsocks is a layer 4 proxy protocol that does not handshake to be anti-censorship. shadowsocks do its own job.

cloudflare aims to proxy HTTP traffic and do its own job.

But when the two mixed-use, it will produce incompatible situation, very helpless but there is no way

PR #2767, opened 2023-05-25
Chore: clean code

savion1024:
Improve code readability.

### DIFF
--- a/component/resolver/resolver.go
+++ b/component/resolver/resolver.go
@@ -141,12 +141,10 @@ func LookupIPWithResolver(ctx context.Context, host string, r Resolver) ([]net.I
 		return []net.IP{node.Data.(net.IP)}, nil
 	}
 
-	if r != nil {
-		if DisableIPv6 {
-			return r.LookupIPv4(ctx, host)
-		}
+	if r != nil && !DisableIPv6 {
 		return r.LookupIP(ctx, host)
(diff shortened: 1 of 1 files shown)


component/resolver/resolver.go
```diff
@@ -141,12 +141,10 @@ func LookupIPWithResolver(ctx context.Context, host string, r Resolver) ([]net.I
 		return []net.IP{node.Data.(net.IP)}, nil
 	}
 
-	if r != nil {
-		if DisableIPv6 {
-			return r.LookupIPv4(ctx, host)
```
Dreamacro, line 146: This is not equivalent, in that `r != nil`, you need to use `r` itself

savion1024: thanks for u reply

savion1024:
Hello,  i think this change may can Improve code readability.Please take a look. thx~

Dreamacro:
I don't think it improves the readability of the code, can you explain? Also the error code from the previous commit is still there.

savion1024:
the logic of both `if` and `else` blocks is to return, it can be replaced with multiple `if` statements, which, in my understanding, can make the code more readable, especially within the `LookupIPWithResolver` method. If you agree with my point of view, I will submit another pull request. This PR will only have one commit. If it's not acceptable, I will close the PR. Thank you for your response

PR #2473, opened 2022-12-22
IPv6 REDIRECT

embeddedlove:
It is OK? Please update the premium core too, or would you please add sniff-tls-sni support for basic version? I need sniffing.

PR #2768, opened 2023-05-25
Add IPv6 support for FreeBSD.

yonas:
Upstream patch from FreeBSD ports which adds IPv6 support.


Dreamacro:
@yonas Are you referring to #2473? But that's Linux, it doesn't necessarily work properly under FreeBSD

yonas:
@Dreamacro I've tested it just now and it works on FreeBSD 13.2.

Dreamacro:
Thanks for the contribution, do you mind if I commit this cherry-pick to premium?

yonas:
@Dreamacro I don't mind if you add it to premium. Thanks for committing.

PR #2778, opened 2023-06-07
d/c/outbound.md: document availablity of interval and url

Artoria2e5:
Uh should I document `lazy` too?


Dreamacro:
> Uh should I document `lazy` too?

If you think it is necessary

Dreamacro:
slate bot did what it had to do, after all it's been a long time since any reply

Artoria2e5:
Again, https://drewdevault.com/2021/10/26/stalebot.html

PR #2788, opened 2023-06-13
Fix: aysnc exchange with new context

kamingchan:
Context may be canceled once caller function return.

e.g.

https://github.com/Dreamacro/clash/blob/dev/dns/resolver.go#L48

https://github.com/Dreamacro/clash/blob/dev/tunnel/tunnel.go#L326


kamingchan:
Updated.

Dreamacro:
thanks for a nice catch 🙌

PR #2795, opened 2023-06-16
Update tun-device.md

jzxo:


Dreamacro:
Can you explain what you are doing?

Dreamacro:
Since the purpose of this pr is not clear and there is no specific explanation, close it first and reopen it if there is a corresponding problem.

PR #2816, opened 2023-06-28
optional provider path

xishang0128:


Dreamacro:
use md5 path is not a good choice

xishang0128:
> use md5 path is not a good choice

How should it be modified? Change provider name?

PR #2818, opened 2023-07-03
Feature: add inbounds for flexible binding inbound

fuyuntt:
related to #2681 
1. [add support for rules based on the local port](https://github.com/Dreamacro/clash/commit/2b06fb83df668c7b106c0c1b043e306e524b81be)
Support for using new rule type: "LOC-PORT", this rule matches the local port being listened on by Clash, e.g. "LOC-PORT,7900,Proxies" indicates that incoming requests from the listening port 7900 will be matched by this rule.
2. [support for listening on multiple extra SOCKS ports](https://github.com/Dreamacro/clash/commit/64ea20b142e6880e4fa9fdb3ccb3c7064283ce84)
Support for listening on multiple additional SOCKS ports, where users can apply 'LOC-PORT' rules to perform different actions on incoming requests to these ports.


Dreamacro:
loc port is a strange name, why not call it Inbound port?

ExtraSocksPort is not generic enough, why can't it support redirect, http, mixed, and so on?

fuyuntt:
1. `Inbound port` is a good name, I will change it to that. 
2. The reason for not supporting all proxy types is that adding ExtraPorts for each proxy type would increase the code size i would have to write. However, since it seems strange to only support multiple Socks ports, I will also add support for other proxy types. 

I will make these changes and commit in the next week.

Dreamacro:
> The reason for not supporting all proxy types is that adding ExtraPorts for each proxy type would increase the code size i would have to write

all listeners impl `C.Listener` so we can use a common structure to manage these listeners. It would be nice if these listeners could specify the binding address in the configuration

fuyuntt:
> > The reason for not supporting all proxy types is that adding ExtraPorts for each proxy type would increase the code size i would have to write
> 
> all listeners impl `C.Listener` so we can use a common structure to manage these listeners. It would be nice if these listeners could specify the binding address in the configuration

good idea, I will try it later.

fuyuntt:
@Dreamacro done

Dreamacro:
Sorry for the late reply.

I noticed a lot of new `ExtraXX` fields that were added to the config, but there were problems with the original port listening: the `bind-address` field controls all inbound bind addresses, which is not flexible enough.

Wouldn't it be much better to add a new field, such as 'inbounds', to declare the address and type of the binding? For example:

```yaml
inbounds:
  - type: socks5
    bind-address: 0.0.0.0:7890
  - type: socks5
    bind-address: 192.168.33.1:7891
  # We can even support one line alias
  - socks5://0.0.0.0:7890
```

fuyuntt:
> Sorry for the late reply.
> 
> I noticed a lot of new `ExtraXX` fields that were added to the config, but there were problems with the original port listening: the `bind-address` field controls all inbound bind addresses, which is not flexible enough.
> 
> Wouldn't it be much better to add a new field, such as 'inbounds', to declare the address and type of the binding? For example:
> 
> ```yaml
> inbounds:
>   - type: socks5
>     bind-address: 0.0.0.0:7890
>   - type: socks5
>     bind-address: 192.168.33.1:7891
>   # We can even support one line alias
>   - socks5://0.0.0.0:7890
> ```
Using new fields might be a good idea, but should we still maintain compatibility with the original fields? If we do maintain compatibility, some issues may be difficult to overcome.
1. what should we do if the configurations in 'inbounds' conflict with the original 'ports' and 'bind-address' configurations?
2. to have the RESTful API return both 'Ports' and 'inbounds' simultaneously, we'd need to store both configurations. This would result in redundant code.

If we don't maintain compatibility, the existing webui would need to be adapted to support the new configuration format.

Dreamacro:
I don't think there are any problems with keeping the current configuration.

> what should we do if the configurations in 'inbounds' conflict with the original 'ports' and 'bind-address' configurations?

do nothing, The only possible problem is if the GUI "adds" the configuration for you and there is a conflict. The old API is just not flexible enough, and people who will use the new inbound field have special needs, so it's not a problem

> to have the RESTful API return both 'Ports' and 'inbounds' simultaneously, we'd need to store both configurations. This would result in redundant code.

For this problem, I haven't any idea how to return it at the moment, but it can be left unresolved in this pr. We just make sure that adding this configuration itself was the right decision

fuyuntt:
Okay, then the RESTful interface will only return the 'inbound' configuration and will not return the 'Ports' configuration under the General setting.  

Another question, if the one-line alias form is used during configuration, the RESTful interface can only return the configuration in its regular form. For example, if you configured it as `socks5://0.0.0.0:7890`, the interface retrieval will return `{"type": "socks5", "bind-address": "0.0.0.0:7890"}`. Is this acceptable?

Dreamacro:
> if you configured it as socks5://0.0.0.0:7890, the RESTful Get will return {"type": "socks5", "bind-address": "0.0.0.0:7890"}. Is this acceptable?

Of course, it could be implemented by a `UnmarshalYAML` struct for both single-line and normal YAML configurations supported

fuyuntt:
Okay, wait my commit.

fuyuntt:
@Dreamacro done.

> Of course, it could be implemented by a UnmarshalYAML struct for both single-line and normal YAML configurations supported

I tried using 'UnmarshalYAML'  but encountered some problems, so I simply used `[]any` to accept the inbounds config and then parsed it to implement the functionality.

Dreamacro:
I pushed a code that solves `UnmarshalYAML`. There's a lot of stuff to change, so the rest of the code needs enough bandwidth to read

component/process/process_windows.go
```diff
@@ -6,9 +6,9 @@ import (
 	"net/netip"
 	"unsafe"
 
-	"golang.org/x/sys/windows"
-
 	"github.com/Dreamacro/clash/common/pool"
+
+	"golang.org/x/sys/windows"
```
Dreamacro, line 11: clash has its own import ordering rules, just run `make lint-fix` to fix the problem here.

constant/listener.go
```diff
@@ -1,7 +1,66 @@
...
+	*i = Inbound(*inner)
+	return nil
+}
+
+func parseInbound(alias string) (*inbound, error) {
+	u, err := url.Parse(alias)
+	if err != nil {
+		return nil, err
+	}
+	listenerType := InboundType(u.Scheme)
```
Dreamacro, line 57: It's better to check the address and scheme here to avoid unnecessary errors.

hub/route/inbounds.go
```diff
@@ -0,0 +1,39 @@
...
+
+	C "github.com/Dreamacro/clash/constant"
+	"github.com/Dreamacro/clash/listener"
+	"github.com/Dreamacro/clash/tunnel"
+
+	"github.com/go-chi/chi/v5"
+	"github.com/go-chi/render"
+)
+
+func inboundRouter() http.Handler {
```
Dreamacro, line 14: Is the implementation of the new api here but breaking the current *-port api inconsistent with the results of our previous discussion?

listener/http/server.go
```diff
@@ -29,11 +29,11 @@ func (l *Listener) Close() error {
 	return l.listener.Close()
 }
 
-func New(addr string, in chan<- C.ConnContext) (*Listener, error) {
+func New(addr string, in chan<- C.ConnContext) (C.Listener, error) {
```
Dreamacro, line 32: return the pointer of the struct is always better than returning the interface

listener/listener.go
```diff
@@ -22,291 +21,115 @@ import (
...
-	defer func() {
-		if err != nil {
-			log.Errorln("Start Mixed(http+socks) server error: %s", err.Error())
-		}
-	}()
-
-	addr := genAddr(bindAddress, port, allowLan)
+func getNeedCloseAndCreateInbound(originInbounds []C.Inbound, newInbounds []C.Inbound) ([]C.Inbound, []C.Inbound) {
+	needCloseMap := map[C.Inbound]bool{}
+	var needClose []C.Inbound
```
Dreamacro, line 104: In clash code, it prefers `variable := []type{}` to avoid returning null when serializing json

test/vmess_test.go
```diff
@@ -18,8 +18,6 @@ func TestClash_Vmess(t *testing.T) {
 	cfg := &container.Config{
 		Image:        ImageVmess,
 		ExposedPorts: defaultExposedPorts,
-		Entrypoint:   []string{"/usr/bin/v2ray"},
```
Dreamacro, line 21: What is the reason for removing these two fields here?

Dreamacro (changes requested):
Except for break and general api, all the other problems are small.

listener/http/server.go
fuyuntt: returning an interface can standardize the return value of each listener's create method, which allows it to be directly referenced in the following code:
```golang
type tcpListenerCreator func(addr string, tcpIn chan<- C.ConnContext) (C.Listener, error)
var tcpListenerCreators = map[C.InboundType]tcpListenerCreator{
	C.InboundTypeHTTP:   http.New,
	C.InboundTypeSocks:  socks.New,
	C.InboundTypeRedir:  redir.New,
	C.InboundTypeTproxy: tproxy.New,
	C.InboundTypeMixed:  mixed.New,
}
```

test/vmess_test.go
fuyuntt: At that time, using the previous configuration, the v2ray container couldn't start. I speculated if the latest version of the V2ray image had changed the Cmd. So, I removed the Cmd configuration to use the Cmd in the image, and it started successfully. However, I am now unable to reproduce this issue. In fact, without the Cmd configuration, it cannot run. So, I will revert the commit from that time.

hub/route/inbounds.go
Dreamacro: The result I expected is to keep all the current api behavior, because it is used by many GUI programs. The new inbound api can only take effect through put config, and other APIs can be implemented later.

test/vmess_test.go
Dreamacro: On my device, deleting these two lines will cause the test to fail.

component/process/process_windows.go
Dreamacro: I reread the sorting rules and you're right. 😂

listener/listener.go
fuyuntt: done

test/vmess_test.go
fuyuntt: done

constant/listener.go
Dreamacro: > It feels a bit heavy to put checks like whether the port is occupied in the unmarshal process.

Just check that the port is filled in and check wether scheme is support

fuyuntt: done

Dreamacro:
The change looks good.

I noticed that `InboundPort` is currently designed to be of type string, but changing it to uint16 is again limited by the current SrcPort and DstPort. 

This is a legacy issue, I'm thinking about revamping the other two into uint16 as well, I'm thinking about forward compatibility, maybe a separate struct type can be used to solve it?

Dreamacro:
But that's beyond the scope of this PR, I'll change it later

fuyuntt:
Why does modifying the data type of the port affect forward compatibility, Isn't it just the type in memory?

Dreamacro:
> Why does modifying the data type of the port affect forward compatibility, Isn't it just the type in memory?

Metadata is serialized to JSON for output to the API, which is used by the GUI.

Dreamacro:
At the beginning of the clash, there weren't many places that used uint16 ports, and std libs like net.SplitHostAndPort led to the use of strings instead of uint16 at the beginning.

fuyuntt:
get

Dreamacro (approved):

fuyuntt:
no problem.

PR #2846, opened 2023-07-25
fix http method in vmess http-opts is not used

sduoduo233:
[https://dreamacro.github.io/clash/configuration/outbound.html#vmess](https://dreamacro.github.io/clash/configuration/outbound.html#vmess)

```yaml
- name: "vmess-http"
  type: vmess
  # interface-name: eth0
  # routing-mark: 1234
  server: server
  port: 443
  uuid: uuid
  alterId: 32
  cipher: auto
  # udp: true
  # network: http
  # http-opts:
  #   # method: "GET"
  #   # path:
  #   #   - '/'
  #   #   - '/video'
  #   # headers:
  #   #   Connection:
  #   #     - keep-alive
```

`method` exists in the config, but it is not used for [creating the http request](https://github.com/Dreamacro/clash/blob/dev/transport/vmess/http.go#L61)


Dreamacro:
Cool, thanks!

PR #2854, opened 2023-07-31
Fix: process resolving unmap addr to handle wrapped IPv4 address

wakou:
I use clashX, I found the PROCESS-NAME rule is invalid after an update. So I use debug to find problem, finally found the function `process find` is refactored, new implement use `net/netip` go module.
In MacOS Monterey, the `SrcIp` property in `metadata` seems  always 16 bit,  when ip is IPv4 address, an IPv6 address wrapping an IPv4 address is actually stored in `metadata`, such as '::ffff:127.0.0.1'. So I added some handling, use `Unmap` method convert  '::ffff:127.0.0.1' to '127.0.0.1'.
I'm not a professional go user, so my changes may need more testing.

### DIFF
--- a/component/process/process_darwin.go
+++ b/component/process/process_darwin.go
@@ -44,7 +44,8 @@ func findProcessPath(network string, from netip.AddrPort, _ netip.AddrPort) (str
 		return "", ErrInvalidNetwork
 	}
 
-	isIPv4 := from.Addr().Is4()
+	fromUnmap := netip.AddrPortFrom(from.Addr().Unmap(), from.Port())
+	isIPv4 := fromUnmap.Addr().Is4()
 
 	value, err := syscall.Sysctl(spath)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Dreamacro:
I'm Ventura 13.5 (22G74) I'm all works great, is the UDP not found?

wakou:
I'm sure it's tcp.
And I read the source code, I found when create a new http or socket connection, the `metadata`'s SrcIp would always a 16 bit from.

Dreamacro:
Is there an easy way I can reproduce it?

wakou:
is there a process-name or process-path rule in your rule set?
switch log-level to debug, then run clash, will print out `process not found` log.
image

Dreamacro:
all `DBG [Matcher] find process success` on my MBP

Dreamacro:
I found that there is something wrong with the open source version. Let me have a look.

Dreamacro:
fixed on 10f4d53

Because all premium uses `netip.Addr`, there is no problem. The open-source version uses legacy `net.IP` which is much more buggy, so that makes the difference

PR #2862, opened 2023-08-05
add PassThrough AdapterType

fuyuntt:
Implemented a new ProxyAdapter named PassThrough. When this ProxyAdapter is matched, no action is performed and the processing proceeds to the next rule. This AdapterType can be used in the following context: Using the SRC-IP-CIDR rule to match a local network host, then matching a select-type ProxyGroup specific to this host. The ProxyGroup includes "REJECT", "DIRECT", "xxxProxy", and "PASS_THROUGH" as options. By choosing the PASS_THROUGH option, the system enables further rule-based routing process.

### DIFF
--- a/adapter/outbound/pass_through.go
+++ b/adapter/outbound/pass_through.go
@@ -1,0 +1,33 @@
+package outbound
+
+import (
+	"context"
+	"errors"
+
+	"github.com/Dreamacro/clash/component/dialer"
+	C "github.com/Dreamacro/clash/constant"
+)
+
+type PassThrough struct {
+	*Base
+}
+
+// DialContext implements C.ProxyAdapter
+func (p *PassThrough) DialContext(ctx context.Context, metadata *C.Metadata, opts ...dialer.Option) (C.Conn, error) {
+	return nil, errors.New("no support")
+}
+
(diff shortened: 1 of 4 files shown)


Dreamacro:
I don't think something like the PassThrough makes sense, it's more like a patch in the existing architecture, which is a very strange thing in terms of the overall structure.

For this kind of requirement, it is solved in premium as follows:
```yaml
rules:
  - if: "in_cidr(src_ip, '192.168.88.100/28')"
    name: XBOX
    rules:
      - DOMAIN-SUFFIX,example.com,DIRECT
      - DOMAIN-SUFFIX,foo.com,DIRECT
  - DOMAIN-SUFFIX,bar.com,proxy
```

fuyuntt:
Indeed, the PassThrough acts as a patch in the current architecture, which can be confusing to use and may cause some unexpected behaviors, such as not being able to terminate existing connections when switching rules.

I'm not familiar with premium, however, based on the configuration you provided, it doesn't seem to solved the requirement I described above. Your example requires specific rule configuration for XBOX, while what I'm wanted is a way to let me choose XBOX whether to REJECT or to use existing rule matching, it's more like some kind of Parental Control Function.

PR #2868, opened 2023-08-09
update requirements

stillbsd:
update requirements and typo


Dreamacro:
clash always updates dependencies before release, unless there is a specific dependency that serves a particular purpose.

And these typos don't look like typos?

BTW, The way to upgrade dependencies is wrong, go.sum is not updated

PR #2878, opened 2023-08-13
Refactor(constant): added rule string enums in configs as RuleConfig for better outside integrations

nekomeowww:
## Summary

There are no exported string type constants for `RuleType` and the rule types from configuration files right now, developers often needs to explicitly write out and enumerate all the possible rule type, and other iota type when integration with clash and process with configuration files, therefore I added the standalone `RuleConfig` type alone with all the currently supported rule types that exported for better API and DX.

## Changes

- Added `RuleConfig` type and all the enums.
- Added compile-time interface check as this is part of the best practice that mentioned by Uber Golang Guide. [References](https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance)
- Added `exhaustive` linter to check whether all the enums are presented in switch branches.

### DIFF
--- a/constant/metadata.go
+++ b/constant/metadata.go
@@ -52,6 +52,8 @@ func (t Type) String() string {
 		return "Redir"
 	case TPROXY:
 		return "TProxy"
+	case TUNNEL:
+		return "Tunnel"
 	default:
 		return "Unknown"
 	}
--- a/constant/rule.go
+++ b/constant/rule.go
@@ -1,5 +1,48 @@
 package constant
 
+// Rule Type String
+const (
+	DomainString        RuleTypeString = "Domain"
+	DomainSuffixString  RuleTypeString = "DomainSuffix"
+	DomainKeywordString RuleTypeString = "DomainKeyword"
+	GEOIPString         RuleTypeString = "GeoIP"
+	IPCIDRString        RuleTypeString = "IPCIDR"
+	SrcIPCIDRString     RuleTypeString = "SrcIPCIDR"
+	SrcPortString       RuleTypeString = "SrcPort"
+	DstPortString       RuleTypeString = "DstPort"
+	InboundPortString   RuleTypeString = "InboundPort"
+	ProcessString       RuleTypeString = "Process"
+	ProcessPathString   RuleTypeString = "ProcessPath"
+	IPSetString         RuleTypeString = "IPSet"
+	MatchString         RuleTypeString = "Match"
+	UnknownString       RuleTypeString = "Unknown"
+)
+
+// Rule Type String represents a rule type, if integrating with configuration files, please use RuleConfigTypeString instead.
+type RuleTypeString string
+
+const (
+	DomainConfigTypeString        RuleConfigTypeString = "DOMAIN"
+	DomainSuffixConfigTypeString  RuleConfigTypeString = "DOMAIN-SUFFIX"
+	DomainKeywordConfigTypeString RuleConfigTypeString = "DOMAIN-KEYWORD"
+	GeoIPConfigTypeString         RuleConfigTypeString = "GEOIP"
+	IPCIDRConfigTypeString        RuleConfigTypeString = "IP-CIDR"
+	IPCIDR6ConfigTypeString       RuleConfigTypeString = "IP-CIDR6"
+	SrcIPCIDRConfigTypeString     RuleConfigTypeString = "SRC-IP-CIDR"
+	SrcPortConfigTypeString       RuleConfigTypeString = "SRC-PORT"
+	DstPortConfigTypeString       RuleConfigTypeString = "DST-PORT"
+	InboundPortConfigTypeString   RuleConfigTypeString = "INBOUND-PORT"
+	ProcessNameConfigTypeString   RuleConfigTypeString = "PROCESS-NAME"
+	ProcessPathConfigTypeString   RuleConfigTypeString = "PROCESS-PATH"
+	IPSetConfigTypeString         RuleConfigTypeString = "IPSET"
+	RuleSetConfigTypeString       RuleConfigTypeString = "RULE-SET"
+	ScriptConfigTypeString        RuleConfigTypeString = "SCRIPT"
+	MatchConfigTypeString         RuleConfigTypeString = "MATCH"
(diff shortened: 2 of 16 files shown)


constant/rule.go
```diff
@@ -19,38 +38,43 @@ const (
...
+		return MATCHString
 	default:
-		return "Unknown"
+		return UnknownString
 	}
 }
 
 type Rule interface {
 	RuleType() RuleType
+	RuleTypeString() RuleTypeString
```
nekomeowww, line 77: Question: Such changed may introduce breaking changes, can be removed, this is not necessary for this change. Should I keep it?

constant/metadata.go
```diff
@@ -52,6 +52,8 @@ func (t Type) String() string {
 		return "Redir"
 	case TPROXY:
 		return "TProxy"
+	case TUNNEL:
```
nekomeowww, line 55: Question: Looks like the `TUNNEL` is missing, should I keep it?

log/log.go
```diff
@@ -79,7 +79,7 @@ func print(data Event) {
 		return
 	}
 
-	switch data.LogLevel {
+	switch data.LogLevel { // nolint:exhaustive
```
nekomeowww, line 82: Question: Looks like the `SLIENT` is missing, should I ignore the linter here?

tunnel/tunnel.go
```diff
@@ -161,7 +161,7 @@ func resolveMetadata(ctx C.PlainContext, metadata *C.Metadata) (proxy C.Proxy, r
 		return
 	}
 
-	switch mode {
+	switch mode { //nolint:exhaustive
```
nekomeowww, line 164: Question: Looks like the `Rule` is missing, should I ignore the linter here?

Dreamacro:
plz base on the dev branch, some changes have been made compared with the master.

nekomeowww:
> plz base on the dev branch, some changes have been made compared with the master.

@Dreamacro rebased and switched the comparing base to `dev`

constant/rule.go
Dreamacro: I don't think it's necessary because there's already a RuleType.

log/log.go
Dreamacro: I think it's a good idea to leave out the nolint annotations as much as possible, here you can just case SLIENT or add a default but don't do anything with it

tunnel/tunnel.go
Dreamacro: You can change `default` to `case Rule` and panic in `default`.

Dreamacro (commented):
Since `RuleTypeString` is only used in one place, I'm not sure what the purpose of having it as a constant. Are you using clash as a library?

nekomeowww:
> Since `RuleTypeString` is only used in one place, I'm not sure what the purpose of having it as a constant. Are you using clash as a library?

Yes. And I didn't find any existing packages that contains the necessary types and constant enums that used by Clash when processing rules and configuration files. Are there any?

Btw, I am planning to add more exported type annotations in the coming PRs where this PR is only focusing on updating the rules related codes and hope this does simplify the review hardness. Is this viable?

Dreamacro:
@nekomeowww 

`RuleTypeString` doesn't seem to be necessary, since you can get the specific string via the String() function: `C.DomainKeyword.String()`, can you describe on the usage scenario?

> Btw, I am planning to add more exported type annotations in the coming PRs where this PR is only focusing on updating the rules related codes and hope this does simplify the review hardness. Is this viable?

it's good, It's painful to put too much change in one PR

nekomeowww:
> @nekomeowww
> 
> `RuleTypeString` doesn't seem to be necessary, since you can get the specific string via the String() function: `C.DomainKeyword.String()`, can you describe on the usage scenario?

You are right. I forgot the `RuleConfigTypeString` was the actual referenced enums when processing the configs and the `RuleTypeString` is probably only useful when deal with logs, etc. I will update the code ASAP to remove the `RuleTypeString` and `Rule.RuleTypeString()` method and simplify the names of type and constants for `RuleConfigTypeString`.

constant/rule.go
nekomeowww: Sure. Removed.

constant/metadata.go
Dreamacro: seems it just missing by mistake

constant/rule.go
```diff
@@ -1,5 +1,27 @@
 package constant
 
+const (
+	DomainTypeString        RuleTypeString = "DOMAIN"
```
Dreamacro, line 4: Is it more friendly to the editor's prompt to put the type in front of it?

When typing `TypeString`, you can see all the values.

btw, isn't RuleTypeString a bit long, is there a better name for it?

nekomeowww: 🤔️ Might be. I could tweak it a little. Since the `Rule` haven't token by the other type yet, is `Rule` a better alternative for it? Or how about `ConfigRule`?

Dreamacro: I have no idea. 🤣

nekomeowww: Updated. Renamed to `RuleConfig`.

Dreamacro: @nekomeowww maybe you forget push?

nekomeowww: > @nekomeowww maybe you forget to push?

Ahh, yes, sorry about that.

Dreamacro (approved):

Dreamacro:
ci failed, seems `exhaustive` not a good lint

nekomeowww:
> ci failed, seems `exhaustive` not a good lint

Should I remove it?

Dreamacro:
@nekomeowww 

I have looked at golangci lint doc, you can add this setting to avoid false positive:

```
linters-settings:
  exhaustive:
    default-signifies-exhaustive: true
```

nekomeowww:
> @nekomeowww
> 
> I have looked at golangci lint doc, you can add this setting to avoid false positive:
> 
> ```
> linters-settings:
>   exhaustive:
>     default-signifies-exhaustive: true
> ```

Updated.

Dreamacro:
Thank you for your patience.

PR #2890, opened 2023-08-19
Feat: embed tzdata for weirdos and newbies

fakeboboliu:
Clash is widely used on various platforms by **_various_** users.

It's handy to embed tzdata for users who are using a weird OS that doesn't have a tzdata and users who are building an Alpine-based Docker image but don't know how to add tzdata for some reason.

## What's the cost

~450 KB


Dreamacro:
should we remove tzdata on https://github.com/Dreamacro/clash/blob/master/Dockerfile?

fakeboboliu:
> should we remove tzdata on https://github.com/Dreamacro/clash/blob/master/Dockerfile?

Yes, or it will increase the image size. I'll commit it to this PR.

PR #2891, opened 2023-08-19
Add support for the VLESS protocol with minimal modifications

crwnet:
Add support for the VLESS protocol, with minimal modifications and reusing as much of the VMESS code as possible to reduce future maintenance effort.


Dreamacro:
First of all, there is nothing wrong with the modified code, but I have two doubts:

* Does it really make sense to add vless if I don't implement features like XTLS?
* If the parameters of vless are not exactly the same as vmess, then writing the code together will not reduce the difficulty of maintenance. Instead, it will increase. They should reuse it at a lower level, instead of mixing the code together. At the user configuration level, they should be split up like shadowsocks and shadowsocksr.

crwnet:
- As far as I am concerned, personally, I don't believe XTLS is a necessary feature (the bandwidth and latency of the network connection between the client and server are much more important than XTLS, and the performance improvement brought by XTLS is imperceptible to the vast majority of users). However, it's regrettable that Clash cannot be used in order to comply with the VLESS protocol specified by certain server-side services.

- The client parameters of the VLESS protocol are a subset of the VMess protocol client parameters. Therefore, users only need to add a parameter "Vless: true" in the proxy configuration to share all the parameters of the VMess protocol, and it supports various TLS+Network combinations (such as vless+ws+tls, vless+h2, vless+grpc, etc.). If support for the quic protocol is added in the future, VLESS will also receive support synchronously without increasing the coding cost. Without planning to support XTLS, I believe this implementation offers the best cost-effectiveness. Otherwise, abstract refactoring and long-term stability testing would be required for the parts shared by VMess and VLESS. Even if branching is needed in the future, handling this part of code is quite straightforward. I think the situation is different from that Shadowsocks and Shadowsocksr, it is reasonable for them to be handled as separate branches because there is very little reusable portion among them.

Dreamacro:
plz add a configuration to test vless on `test/`

crwnet:
Test cases based on the VMess test have been added, and it have successfully passed all communication scenarios defined by the VMess.

Dreamacro:
The vless protocol itself can be used without any encryption. I'm concerned about whether this could trigger misuse.

I know v2ray supports this, but too much flexibility is sometimes not good. Even clash wants to avoid unnecessary misuse, like vless with nothing encrypted.

crwnet:
I believe that security is a concern that service providers must consider, while clients should focus on maximizing usability. Each communication protocol may have its own unique use cases, just like the HTTP protocol still serves a purpose today. We only need to indicate in the documentation which communication combinations are unencrypted, so that users are aware.

adapter/outbound/vmess.go
```diff
@@ -263,15 +263,19 @@ func (v *Vmess) ListenPacketContext(ctx context.Context, metadata *C.Metadata, o
 	return newPacketConn(&vmessPacketConn{Conn: c, rAddr: metadata.UDPAddr()}, v), nil
 }
 
-func NewVmess(option VmessOption) (*Vmess, error) {
+func NewVmess(option VmessOption, isVless bool) (*Vmess, error) {
```
Dreamacro, line 266: I prefer to turn this NewVmess into `newVmess` and add two new `NewVmess` and `NewVless`

Dreamacro: and don't forget add vless proxy type

crwnet: > I prefer to turn this NewVmess into `newVmess` and add two new `NewVmess` and `NewVless`

ok let's do it

crwnet: > and don't forget add vless proxy type

Do you mean "type: vless" profile parsing? it's already there.

Dreamacro: > Do you mean "type: vless" profile parsing? it's already there.

`Base` struct has the `tp` field, it would return for API

crwnet: Ok. Despite the small amount of code, these modifications have succeeded in making users perceive VLESS as a fully standalone implementation rather than the originally intended appendix to VMESS. Moreover, it have prepared for future compatibility, which is indeed a improvement.

Dreamacro:
Thanks for the contribution, would you mind cherry-picking it into premium?

crwnet:
> Thanks for the contribution, would you mind cherry-picking it into premium?

You're welcome! Sure, I don't mind.

PR #2907, opened 2023-08-27
fix: exempt version api access

septs:
exempt read `/version` (build hash) and `/` (hello response), needed authenticated


Dreamacro:
It is intentional, many people put the clash api on the public web (though I don't know why), and putting API without authentication actually increases the risks of detection.

PR #2908, opened 2023-08-27
feat: add reload signal support

septs:
Provide reload configuration configuration capability

e.g:

`systemctl reload clash` (systemd)
see https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecReload=

OR

`service clash reload` (procd)
see https://openwrt.org/docs/guide-developer/procd-init-scripts#reloading_service_setup


Dreamacro:
If the user has used the API to make configuration changes themselves (which is mostly what is done in the GUI), then this reload behavior becomes weird because it always reloads from the default configuration.

septs:
I'm developing a managed clash service, need the capability

Dreamacro:
It will make `. /clash -f /path/to/config` invalidated and load wrong `~/.config/config.yaml`

main.go
```diff
@@ -100,7 +100,22 @@ func main() {
 		log.Fatalln("Parse config error: %s", err.Error())
 	}
 
+	go onReloadSignal()
+
 	sigCh := make(chan os.Signal, 1)
 	signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
 	<-sigCh
```
Dreamacro, line 107: There is no need to start a new goroutine, just select it together with the term signal.

```go
for {
    select {
        case <-termSign:
            return
        case <-hupSign:
            // do something
    }
}
```

main.go
```diff
@@ -100,7 +100,18 @@ func main() {
...
+	termSign := make(chan os.Signal, 1)
+	hupSign := make(chan os.Signal)
+	signal.Notify(termSign, syscall.SIGINT, syscall.SIGTERM)
+	signal.Notify(hupSign, syscall.SIGHUP)
+	for {
+		select {
+		case <-termSign:
+			return
+		case <-hupSign:
+			if cfg, err := executor.ParseWithPath(C.Path.Config()); err == nil {
```
Dreamacro, line 112: Should we log on err?

main.go
```diff
@@ -100,7 +100,20 @@ func main() {
...
+				executor.ApplyConfig(cfg, true)
+			} else {
+				log.Fatalln("Parse config error: %s", err.Error())
```
Dreamacro, line 115: `Fatalln` would exit clash, maybe just Errorln?

PR #2913, opened 2023-08-28
feat: add xdg base support

### LINKED ISSUE #2910: [Feature] XDG Base Directory Support
### Prerequisites

- [X] I understand that this is the official open-source version of the Clash core, **only providing support for the open-source version or Premium version**
- [X] I have looked for my idea in [the issue tracker](https://github.com/Dreamacro/clash/issues?q=is%3Aissue+label%3Aenhancement), **and found none of which being related**
- [X] I have read the [official Wiki](https://dreamacro.github.io/clash/)

### Description

The clash not support, use xdg base directory redirect dot-config path

https://github.com/Dreamacro/clash/blob/e5f2396f810a778792cc305d3cee293b5a677c25/constant/path.go#L13-L21

XDG provide the variable can use user-defined dot-config to new location

> **XDG_CONFIG_HOME**
>
> Where user-specific configurations should be written (analogous to `/etc`).
> Should default to `$HOME/.config`.

more see https://wiki.archlinux.org/title/XDG_Base_Directory

### Possible Solution

https://wiki.archlinux.org/title/XDG_Base_Directory
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

septs:
closes #2910


septs:
`os.UserConfigDir()` supports XDG Base, see https://github.com/golang/go/blob/071aed2aaa0ed819582c5bff44b70d43c61f504a/src/os/file.go#L485-L493

Dreamacro:
I don't think UserConfigDir's behavior is suitable for macOS. I'll use my mbp as an example.

On `$HOME/Library/Application Support`, There are many folders of GUI programs, and the only CLI programs are go and zoxide.

on `$HOME/.config`, like gh, iterm2, htop, wireshark, alacritty, starship etc. It seems that CLI programs prefer to put it in `$HOME/.config`.

Dreamacro:
Thanks a lot

PR #2915, opened 2023-08-31
Fix: should check all ips need to fallback

yaling888:


Dreamacro:
Have you encountered any special use cases?

Dreamacro:
thanks

PR #2930, opened 2023-09-11
feat: add PROTOCOL rule support

septs:
Use Golang builtin protocol lookup

see https://pkg.go.dev/net#LookupPort
see https://github.com/golang/go/blob/09465129bd9adc2713c16656ac614df396117355/src/net/lookup.go#L33-L58


septs:
for Unix systems, golang use `getaddrinfo` function get port by service name

https://github.com/golang/go/blob/06138963a2a16e0ac982ac85877092920a21e60f/src/net/cgo_unix.go#L115
https://man7.org/linux/man-pages/man3/getaddrinfo.3.html
https://manpages.debian.org/bullseye/manpages/services.5.en.html

for Windows, golang use `GetAddrInfoW` function get port by service name

https://github.com/golang/go/blob/06138963a2a16e0ac982ac85877092920a21e60f/src/net/lookup_windows.go#L215C15-L215C27
https://learn.microsoft.com/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfow

septs:
Typical application:

## Goal

Exempt SSH, SMTP, etc protocol

## Sample Rule

```
PROTOCOL,ssh,DIRECT
PROTOCOL,smtp,DIRECT
PROTOCOL,imaps,DIRECT
PROTOCOL,pop3s,DIRECT
```

## Reference

Some service providers block the SSH, SMTP, etc protocol

> 22端口已被封禁，以及一些高危端口，以保障服务器的安全和正常运行。

from https://doc.boccc.co/zh/article/56ab5q2i55qe5yqf6io95lul5yk572r56uz-4vccag/

constant/rule.go
```diff
@@ -11,6 +11,7 @@ const (
 	RuleConfigSrcPort       RuleConfig = "SRC-PORT"
 	RuleConfigDstPort       RuleConfig = "DST-PORT"
 	RuleConfigInboundPort   RuleConfig = "INBOUND-PORT"
+	RuleConfigProtocolPort  RuleConfig = "PROTOCOL"
```
nekomeowww, line 14: Should it be `PROTOCOL-PORT` that matched up the pattern with existing rules?

septs: or `DST-PORT` provide use service-name capability

```
DST-PORT,ssh,DIRECT
```

septs: ## PROTOCOL

Rule matches if the protocol of the request matches. The possible values are HTTP, HTTPS, TCP, UDP, DOH, DOH3, DOQ.

```
PROTOCOL,HTTP,DIRECT
```

more see https://manual.nssurge.com/rule/misc-rule.html

PR #2936, opened 2023-09-17
feat: add Surge-compatible `DEST-PORT` rule

septs:
## Surge

> **DEST-PORT**
>
> Rule matches if the target port of the request matches.
>
> ```
> DEST-PORT,80,DIRECT

from https://manual.nssurge.com/rule/misc-rule.html

## Clash

> **DST-PORT**
>
> DST-PORT rules are used to route packets based on the destination port of the packet.
>
> `DST-PORT,80,policy` routes any packets to the port 80 to `policy`.

from https://dreamacro.github.io/clash/configuration/rules.html#dst-port


Dreamacro:
@septs Can you tell me why you turned these pr's on and off?

PR #2954, opened 2023-10-08
Feature: nameserver support tcp query over socks5 to avoid pollution attacks

qiuyi21:
nameserver support tcp query over socks5 to avoid pollution attacks, configure like this:

dns:
  enable: true
  listen: :53
  nameserver:
    - 'socks5://127.0.0.1:7891#8.8.8.8:53'     # tcp query 8.8.8.8:53 via socks5://127.0.0.1:7891


InspoOnU:
Clash still resolves DNS locally when encountering UDP as a domain, so I think it's necessary.

